PR #18, opened 2021-07-27
Override authority when building mTLS and TLS channels.

wangyaopw:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/18)

### DIFF
--- a/src/main/kotlin/org/wfanet/measurement/common/grpc/Channel.kt
+++ b/src/main/kotlin/org/wfanet/measurement/common/grpc/Channel.kt
@@ -39,9 +39,21 @@ fun buildPlaintextChannel(target: String): ManagedChannel {
  * @param target the URI or authority string for the target server.
  * @param clientCerts the collection of client certificate and private key, as well as the trusted
  * server certificates.
+ * @param hostName the expected DNS hostname from the Subject Alternative Name (SAN) of the server's
+ * certificate
  */
-fun buildMutualTlsChannel(target: String, clientCerts: SigningCerts): ManagedChannel {
-  return NettyChannelBuilder.forTarget(target).sslContext(clientCerts.toClientTlsContext()).build()
+fun buildMutualTlsChannel(
+  target: String,
+  clientCerts: SigningCerts,
+  hostName: String? = null
+): ManagedChannel {
+  val channelBuilder =
+    NettyChannelBuilder.forTarget(target).sslContext(clientCerts.toClientTlsContext())
+  return if (hostName == null) {
+    channelBuilder.build()
+  } else {
+    channelBuilder.overrideAuthority(hostName).build()
+  }
 }
 
 /**
@@ -49,14 +61,21 @@ fun buildMutualTlsChannel(target: String, clientCerts: SigningCerts): ManagedCha
  *
  * @param target the URI or authority string for the target server
  * @param trustedServerCerts trusted server certificates.
+ * @param hostName the expected DNS hostname from the Subject Alternative Name (SAN) of the server's
+ * certificate
  */
 fun buildTlsChannel(
   target: String,
-  trustedServerCerts: Collection<X509Certificate>
+  trustedServerCerts: Collection<X509Certificate>,
+  hostName: String? = null
 ): ManagedChannel {
-  return NettyChannelBuilder.forTarget(target)
-    .sslContext(trustedServerCerts.toClientTlsContext())
-    .build()
+  val channelBuilder =
+    NettyChannelBuilder.forTarget(target).sslContext(trustedServerCerts.toClientTlsContext())
+  return if (hostName == null) {
+    channelBuilder.build()
+  } else {
+    channelBuilder.overrideAuthority(hostName).build()
+  }
 }
 
 /** Add shutdownHook to a managedChannel */


oliver-amzn (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/18)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @SanjayVas and @wangyaopw)

___
*[src/main/kotlin/org/wfanet/measurement/common/grpc/Channel.kt, line 46 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/18#-MfdkckiAvirpL6RAWwG:-MfdkckiAvirpL6RAWwH:b-smhtuz) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/eb426dc19733d661a87521405470bd153e8dcde7/src/main/kotlin/org/wfanet/measurement/common/grpc/Channel.kt#L46)):*
> ```
> overrideAuthority
> ```

What does this mean, to override the authority? 
Do you maybe have a link that explains this?

wangyaopw (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/18)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @oliver-amzn and @SanjayVas)

___
*[src/main/kotlin/org/wfanet/measurement/common/grpc/Channel.kt, line 46 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/18#-MfdkckiAvirpL6RAWwG:-MfdnheB8zLCyeYMrJqX:boxbj89) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/eb426dc19733d661a87521405470bd153e8dcde7/src/main/kotlin/org/wfanet/measurement/common/grpc/Channel.kt#L46)):*
<details><summary><i>Previously, oliver-amzn (Oliver) wrote…</i></summary><blockquote>

> ```
> overrideAuthority
> ```

What does this mean, to override the authority? 
Do you maybe have a link that explains this?
</blockquote></details>

https://grpc.github.io/grpc-java/javadoc/io/grpc/ManagedChannelBuilder.html#overrideAuthority-java.lang.String-

Quoting Sanjay's message left somewhere else.
"
Summary of the issue: When you generate a cert, you specify the hostname and/or IP in the subject alternative name (SAN). If you have dynamic IPs (such as those assigned by K8s when bringing things up), then you can't have the correct one specified in the cert. Therefore, you need a way to tell the client what you actually expect the SAN to be instead of using the real host/IP of the server.

There are two options for this: configure custom DNS in your K8s environment such that you map the SAN hostname to the dynamic IP, or just use overrideAuthority.
"

oliver-amzn (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/18)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @SanjayVas)

___
*[src/main/kotlin/org/wfanet/measurement/common/grpc/Channel.kt, line 46 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/18#-MfdkckiAvirpL6RAWwG:-Mfdqg2DBI1aOtZzhN3O:b-408r8m) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/eb426dc19733d661a87521405470bd153e8dcde7/src/main/kotlin/org/wfanet/measurement/common/grpc/Channel.kt#L46)):*
<details><summary><i>Previously, wangyaopw (Yao Wang) wrote…</i></summary><blockquote>

https://grpc.github.io/grpc-java/javadoc/io/grpc/ManagedChannelBuilder.html#overrideAuthority-java.lang.String-

Quoting Sanjay's message left somewhere else.
"
Summary of the issue: When you generate a cert, you specify the hostname and/or IP in the subject alternative name (SAN). If you have dynamic IPs (such as those assigned by K8s when bringing things up), then you can't have the correct one specified in the cert. Therefore, you need a way to tell the client what you actually expect the SAN to be instead of using the real host/IP of the server.

There are two options for this: configure custom DNS in your K8s environment such that you map the SAN hostname to the dynamic IP, or just use overrideAuthority.
"
</blockquote></details>

Ah right, that sounds familiar now, thanks for quote and link!

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/18)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @wangyaopw)

___
*[src/main/kotlin/org/wfanet/measurement/common/grpc/Channel.kt, line 46 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/18#-MfeECMO0rY914pZUF4s:-MfeECMPBqNoAP4gcXI1:b34bya6) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/eb426dc19733d661a87521405470bd153e8dcde7/src/main/kotlin/org/wfanet/measurement/common/grpc/Channel.kt#L46)):*
> ```
> "localhost"
> ```
Take this in as a method parameter, documenting that this is the expected DNS hostname from the Subject Alternative Name (SAN) of the server's certificate.

We'd need to configure these using a flag that goes with each service target flag.

wangyaopw (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/18)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[src/main/kotlin/org/wfanet/measurement/common/grpc/Channel.kt, line 46 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/18#-MfeECMO0rY914pZUF4s:-Mfi0fPa5Yr4K1fbWwBg:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/eb426dc19733d661a87521405470bd153e8dcde7/src/main/kotlin/org/wfanet/measurement/common/grpc/Channel.kt#L46)):*

> ```
> "localhost"
> ```
Take this in as a method parameter, documenting that this is the expected DNS hostname from the Subject Alternative Name (SAN) of the server's certificate.

We'd need to configure these using a flag that goes with each service target flag.
</blockquote></details>

Done.

SanjayVas (approved):
> Reviewed 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/18)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @wangyaopw)

PR #45, opened 2021-10-06
Implement AeadStore for AEAD Blob/Object Storage

corbantek:
Aead/AeadTink:
 - Aead is a ByteString enabled implementation of tink.Aead, but also essentially wraps tink.Aead as not to expose Tink to callers (or if another Aead impl is used in the future)
 - AeadTink is a Tink specific implemenation for Aead and shoudl be used in conjunction with Tink/Integrations (GCP/AWS)

Store:
 - `open` methods get/write for Store to allow AeadStore to encrypt/decrypte when writing/reading blobs

AeadStore:
 - An abstract AeadStore that encrypts/decrypts before writing/reading to Blob/Object Storage
 - AeadStore should be extended per use (IE ElGamalKeyAeadStore...)

Misc:
 - Move Tink import to common-jvm


stevenwarejones (changes requested):
> Reviewed 8 of 8 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/45)* status: all files reviewed, 5 unresolved discussions (waiting on @corbantek, @duliomatos1, @efoxepstein, @SanjayVas, and @wangyaopw)

___
*[src/main/kotlin/org/wfanet/measurement/common/crypto/TinkAead.kt, line 37 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/45#-MlLIrqG2FyZ1opg1tEP:-MlLIrqG2FyZ1opg1tEQ:b636ciq) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/12e380a9005c91310220930fece4f0228267e37f/src/main/kotlin/org/wfanet/measurement/common/crypto/TinkAead.kt#L37)):*
> ```Kotlin
>    * @return resulting ciphertext
>    */
>   override fun encrypt(plaintext: ByteArray?, associatedData: ByteArray?): ByteArray {
> ```

not sure why you allow nullable values here, esp for the ciphertext. If they have a null associatedData they can just call `encrypt(plaintext)`
___
*[src/main/kotlin/org/wfanet/measurement/common/crypto/TinkAead.kt, line 55 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/45#-MlLIyql6I3JeAv_-zoM:-MlLIyql6I3JeAv_-zoN:b-7ropln) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/12e380a9005c91310220930fece4f0228267e37f/src/main/kotlin/org/wfanet/measurement/common/crypto/TinkAead.kt#L55)):*
> ```Kotlin
>    * @return resulting plaintext
>    */
>   override fun decrypt(ciphertext: ByteArray?, associatedData: ByteArray?): ByteArray {
> ```

ditto
___
*[src/main/kotlin/org/wfanet/measurement/storage/AeadStore.kt, line 67 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/45#-MlLKA7D5Tod7R6v6PLE:-MlLKA7D5Tod7R6v6PLF:b-grn421) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/12e380a9005c91310220930fece4f0228267e37f/src/main/kotlin/org/wfanet/measurement/storage/AeadStore.kt#L67)):*
> ```Kotlin
>   /** A blob that will decrypt the content when read */
>   inner class AeadBlob(val blob: Blob) : StorageClient.Blob {
>     override val size: Long = readEncyptedBytes().size().toLong()
> ```

not: `readEncryptedBytes`
___
*[src/main/kotlin/org/wfanet/measurement/storage/AeadStore.kt, line 79 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/45#-MlLKC0M2pTxHhjrO1XK:-MlLKC0M2pTxHhjrO1XL:b-7ropln) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/12e380a9005c91310220930fece4f0228267e37f/src/main/kotlin/org/wfanet/measurement/storage/AeadStore.kt#L79)):*
> ```Kotlin
>     override fun delete() = blob.delete()
> 
>     private fun readEncyptedBytes(): ByteString {
> ```

ditto
___
*[src/main/kotlin/org/wfanet/measurement/storage/BUILD.bazel, line 30 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/45#-MlLJZ3X4vF076F3IQ7k:-MlLJZ3X4vF076F3IQ7l:bad6t7j) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/12e380a9005c91310220930fece4f0228267e37f/src/main/kotlin/org/wfanet/measurement/storage/BUILD.bazel#L30)):*
> ```Starlark
> kt_jvm_library(
>     name = "aead_store",
>     srcs = ["AeadStore.kt"],
> ```

why not just bundle this into the existing `store` dependency?

corbantek (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/45)* status: all files reviewed, 14 unresolved discussions (waiting on @corbantek, @duliomatos1, @efoxepstein, @SanjayVas, @stevenwarejones, and @wangyaopw)

___
*[src/main/kotlin/org/wfanet/measurement/common/crypto/Aead.kt, line 21 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/45#-MlLZBiI7wf_6C9wyS9Z:-MlLk8SDC6OoU5kTDizs:bb74ce2) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/12e380a9005c91310220930fece4f0228267e37f/src/main/kotlin/org/wfanet/measurement/common/crypto/Aead.kt#L21)):*
<details><summary><i>Previously, efoxepstein (Eli Fox-Epstein) wrote…</i></summary><blockquote>

Maybe import Tink's Aead with an alias so this is less confusing-looking?
</blockquote></details>

Sure
___
*[src/main/kotlin/org/wfanet/measurement/common/crypto/Aead.kt, line 21 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/45#-MlLZTiB30Tyh-uWAm6Q:-MlLkA9U4WO2EgxULIss:b-yrmje0) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/12e380a9005c91310220930fece4f0228267e37f/src/main/kotlin/org/wfanet/measurement/common/crypto/Aead.kt#L21)):*
<details><summary><i>Previously, efoxepstein (Eli Fox-Epstein) wrote…</i></summary><blockquote>

Do we need tink.Aead to be a superinterface? Or can we just have the interface be these two functions?
</blockquote></details>

Unfortunately we do because this is used by consent-signalling and inorder to do private key writing (without using ClearTextWriter), it requires ONLY the Tink.Aead interface,  so inorder to reuse this object later, I need to have it be the super-interface.

Essentially I'm addressing this comment https://reviewable.io/reviews/world-federation-of-advertisers/consent-signaling-client/25#-MlGcEk249y5LoR09BEi with this.  I will push an updated version of consent-signalling soon if that help, but I'm temporarily working by just using this commit.
___
*[src/main/kotlin/org/wfanet/measurement/common/crypto/TinkAead.kt, line 24 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/45#-MlLOkWR4yEeDDAWk1OU:-MlLlIX8-O6nNxS4oc5D:b53iluf) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/12e380a9005c91310220930fece4f0228267e37f/src/main/kotlin/org/wfanet/measurement/common/crypto/TinkAead.kt#L24)):*

can't you just pass in null as the associatedData? the documentation suggests they treat zero-length byte arrays as equivalent to null.
</blockquote></details>

I thought its always preferred in Kotlin to not use null (really ever)
___
*[src/main/kotlin/org/wfanet/measurement/common/crypto/TinkAead.kt, line 38 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/45#-MlLNf6xAMZ4c99mjKwc:-MlLlPfF3q6I_FPONOIk:bmgcut4) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/12e380a9005c91310220930fece4f0228267e37f/src/main/kotlin/org/wfanet/measurement/common/crypto/TinkAead.kt#L38)):*

do you need the second `requireNotNull` I thought tink allowed null associatedData as you state in the comments.
</blockquote></details>

I can remove, I just figured if we are in Kotlin I'd force everything to not be null
___
*[src/main/kotlin/org/wfanet/measurement/storage/AeadStore.kt, line 48 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/45#-MlL_jbgEN2lIwnnBbl0:-MlLlrNNDYBj06DRrFqM:b-q9t4dv) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/12e380a9005c91310220930fece4f0228267e37f/src/main/kotlin/org/wfanet/measurement/storage/AeadStore.kt#L48)):*
<details><summary><i>Previously, efoxepstein (Eli Fox-Epstein) wrote…</i></summary><blockquote>

General question to which I'm not sure what the right answer is: what about not making `write` open and instead adding an abstract `transform` function to `Store` that's called on the Flow within `write`?
</blockquote></details>

Thats a great idea.  Would you want it to be abstract or just be an open method?  If its abstract then all callers will have to implement `transform` (but probably only 1% need it)
___
*[src/main/kotlin/org/wfanet/measurement/storage/AeadStore.kt, line 48 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/45#-MlLaC-e-wB6slDOV7Jx:-MlLmDH-CEzcD_2nb9Zm:bk5admf) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/12e380a9005c91310220930fece4f0228267e37f/src/main/kotlin/org/wfanet/measurement/storage/AeadStore.kt#L48)):*
<details><summary><i>Previously, efoxepstein (Eli Fox-Epstein) wrote…</i></summary><blockquote>

Should we use streaming AEAD instead so we don't need to read the entire `content` flow at once?
</blockquote></details>

Thats a good suggestion but I'll need to see how that works with the KmsClient from Tink.  Can we make this a TODO or a follow up task?
___
*[src/main/kotlin/org/wfanet/measurement/storage/AeadStore.kt, line 67 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/45#-MlLZs_aEyqolG115dRf:-MlLnReE2coN5mv-mt3a:bebrt1w) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/12e380a9005c91310220930fece4f0228267e37f/src/main/kotlin/org/wfanet/measurement/storage/AeadStore.kt#L67)):*
<details><summary><i>Previously, efoxepstein (Eli Fox-Epstein) wrote…</i></summary><blockquote>

Should this be `by lazy`?
</blockquote></details>

Will do
___
*[src/main/kotlin/org/wfanet/measurement/storage/AeadStore.kt, line 84 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/45#-MlLZyLZ7YBUhk_k7OaQ:-MlLn_F_EQ8XbrmyJJUZ:b-oi5emx) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/12e380a9005c91310220930fece4f0228267e37f/src/main/kotlin/org/wfanet/measurement/storage/AeadStore.kt#L84)):*
<details><summary><i>Previously, efoxepstein (Eli Fox-Epstein) wrote…</i></summary><blockquote>

Probably want to cache the result of this (maybe just put it in a `by lazy` private variable instead of in a function.

```
val encryptedBytes: ByteString by lazy {
  runBlocking { blob.read(...) }
}
```
</blockquote></details>

Cool, didn't know this syntax

Will do
___
*[src/main/kotlin/org/wfanet/measurement/storage/BUILD.bazel, line 30 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/45#-MlLJZ3X4vF076F3IQ7k:-MlLnizWDqEJ7xbFlT_f:b30enx7) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/12e380a9005c91310220930fece4f0228267e37f/src/main/kotlin/org/wfanet/measurement/storage/BUILD.bazel#L30)):*

why not just bundle this into the existing `store` dependency?
</blockquote></details>

Wasn't sure if everyone wanted AeadStore with store, but I can combine

efoxepstein (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/45)* status: all files reviewed, 15 unresolved discussions (waiting on @corbantek, @duliomatos1, @efoxepstein, @SanjayVas, @stevenwarejones, and @wangyaopw)

___
*[src/main/kotlin/org/wfanet/measurement/storage/AeadStore.kt, line 48 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/45#-MlLaC-e-wB6slDOV7Jx:-MlLwitXDZ-3SMJ7_Cm5:b-vcm9kl) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/12e380a9005c91310220930fece4f0228267e37f/src/main/kotlin/org/wfanet/measurement/storage/AeadStore.kt#L48)):*

I don't think this is possible. This is calling out to KMS, which doesn't support streaming AEAD.
</blockquote></details>

Sounds good, marking resolved

corbantek (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/45)* status: all files reviewed, 15 unresolved discussions (waiting on @corbantek, @duliomatos1, @efoxepstein, @SanjayVas, @stevenwarejones, and @wangyaopw)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/45#-MlLojU9BfuKZFgs1q3o:-MlLuSCB3rvJRoTuba3h:b-izc9dc) (no related file):*

As alluded to in world-federation-of-advertisers/consent-signaling-client#25 and in offline discussion, I'm still not sure this is the right abstraction for both Tink and Storage.

1. Each Store is really just meant to map a given context to a privately-prefixed blob key. StorageClient is really the thing that handles shuttling the bytes around.  It seems that it would be simpler to create a KmsEncryptedStorageClient that wraps a given StorageClient instance. Then you could use whatever StorageClient you want underneath it, and whatever Store you want on top.
2. Unless we specifically need an Aead primitive abstraction for other reasons, we could likely just have a Tink-specific KmsEncryptedStorageClient. The Tink crypto factory method would be something like `makeKmsStorageClient(storageClient: StorageClient, keyUri: String): StorageClient`
</blockquote></details>

For
1. I understand, but I was looking at it from the perspective of being able to use different StorageClients with my AeadStore.  I could easily instantiate a AeadStore(InMemoryStorageClient) for testing.  Also this also makes it easy for any current Store impls to switch to Aead by just extending AeadStore instead of Store

2. I originally wanted to do this, but Tink deprecated all of their `keyUri` methods and really wants you to use their `register` that automatically generates an AWS or GCP KmsClient, that ends up returning the Aead interface directly.  If we deal with keyUri's in our code we will just be reimplementing a lot of the integration code in Tink  

---
___
*[src/main/kotlin/org/wfanet/measurement/common/crypto/Aead.kt, line 21 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/45#-MlLZTiB30Tyh-uWAm6Q:-MlLsuoRCcJKPMTk3vGI:b-ow4u27) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/12e380a9005c91310220930fece4f0228267e37f/src/main/kotlin/org/wfanet/measurement/common/crypto/Aead.kt#L21)):*

I would state more unequivocally that this is part of the public interface that's supposed to abstract Tink, so should not reference it publicly.
</blockquote></details>

I could, but I pass around this Aead interface and without implementing here then I can't use this interface inside CMM consent-signalling and have to use the TinkImpl directly, which I'm trying to avoid...

I understand we don't want to expose Tink which I dont think we are as its org.wfanet.wfameasurement.common.crypto.Aead.  If I don't do this I will have to use the ClearTextWriter again) 
___
*[src/main/kotlin/org/wfanet/measurement/common/crypto/TinkAead.kt, line 17 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/45#-MlLn-e96ULWkHHcYa60:-MlLwdr3Dmkf7TmitZpG:bhy9q7g) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/12e380a9005c91310220930fece4f0228267e37f/src/main/kotlin/org/wfanet/measurement/common/crypto/TinkAead.kt#L17)):*

> ```
> TinkAead
> ```
Similarly to the Aead interface, it's really confusing to have two different symbols in the same file with the same name.
</blockquote></details>

Will fix

corbantek (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/45)* status: all files reviewed, 15 unresolved discussions (waiting on @corbantek, @duliomatos1, @efoxepstein, @SanjayVas, @stevenwarejones, and @wangyaopw)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/45#-MlLojU9BfuKZFgs1q3o:-MlM7OVt92zk23beXT2f:bqf2ihb) (no related file):*

1. This still allows the use of different StorageClients as it just wraps an existing one. The point is just the separation of concerns: StorageClient cares about shuttling bytes, and Store cares about generating blob keys. Encryption is a feature on the bytes, not on the blob key. Note that this is a distinctly different case than something that's concerned with converting encrypted bytes to/from `KeysetHandle` instances, which should be read from a Store that doesn't do the decryption on its own.
2. I'm not sure what you mean here. Wouldn't the impl be something like what is below (assuming on startup you use config flags to determine which KMS client type to register)? I don't think these methods are deprecated.
```
/** Returns a wrapper [StorageClient] that encrypts/decrypts blob data using the specified KMS [keyUri]. */
fun makeKmsStorageClient(storageClient: StorageClient, keyUri: String): StorageClient {
  val aead = KmsClients.get(keyUri).getAead(keyUri)
  return TinkKmsStorageClient(storageClient, aead)
}
```
</blockquote></details>

Okay I'll restructure this PR to this design tonight

---

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/45)* status: 4 of 9 files reviewed, 15 unresolved discussions (waiting on @corbantek, @duliomatos1, @efoxepstein, @SanjayVas, @stevenwarejones, and @wangyaopw)

___
*[src/main/kotlin/org/wfanet/measurement/common/crypto/Aead.kt, line 21 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/45#-MlLZTiB30Tyh-uWAm6Q:-MlRGv62ED_OfltxyMSB:bbjb71f) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/12e380a9005c91310220930fece4f0228267e37f/src/main/kotlin/org/wfanet/measurement/common/crypto/Aead.kt#L21)):*
> I understand we don't want to expose Tink which I don't think we are as its org.wfanet.wfameasurement.common.crypto.Aead.

This is definitely still exposing it. Code that depends on this interface will have to also depend on Tink, regardless of whether they're using a Tink implementation of this interface.

> I could, but I pass around this Aead interface and without implementing here then I can't use this interface inside CMM consent-signalling and have to use the TinkImpl directly, which I'm trying to avoid...

Do we even need the Aead primitive exposed? Meaning do we even need an `AeadStorageClient` in common JVM? We can just use StorageClient itself as the abstraction, and a `TinkKmsStorageClient`. We're not necessarily trying to optimize for sharing code with a non-Tink implementation, but rather just making sure that the seam between the rest of the system and Tink is well-defined.

Recall that there should be a single entry point in the system that specifies that it's using the Tink implementations and the KMS key URI (e.g. a server's `main` function). The design of this is coupled with how you abstract Tink. For example, consider a `CryptoProvider` interface with factory methods such as `makeKmsStorageClient` which returns a `StorageClient` instance. Then a `TinkCryptoProvider` that implements this interface, and has its`makeKmsStorageClient` return a `TinkKmsStorageClient`. The entry point then passes this along as a dependency to the relevant top-level constructors. I think it would be worth taking a step back and having a design for this laid out so it can be discussed without the PR code churn.

stevenwarejones (approved):
> Reviewed 5 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/45)* status: all files reviewed, 8 unresolved discussions (waiting on @corbantek, @duliomatos1, @efoxepstein, @SanjayVas, and @wangyaopw)

corbantek (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/45)* status: all files reviewed, 3 unresolved discussions (waiting on @duliomatos1, @efoxepstein, @SanjayVas, and @wangyaopw)

___
*[src/main/kotlin/org/wfanet/measurement/common/crypto/Aead.kt, line 21 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/45#-MlLZTiB30Tyh-uWAm6Q:-MlTETtiBnsyLeYLer3p:bfqsvby) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/12e380a9005c91310220930fece4f0228267e37f/src/main/kotlin/org/wfanet/measurement/common/crypto/Aead.kt#L21)):*
<details><summary><i>Previously, efoxepstein (Eli Fox-Epstein) wrote…</i></summary><blockquote>

@corbantek -- here's my thinking that'll get this unblocked ASAP. What Panel Match needs by a week ago is `TinkKmsStorageClient` (and some factory to create it and return it as a `StorageClient`).

The rest of the pieces are less urgent, so we can take some time to think about how they're used.

Do you think you could just implement `TinkKmsStorageClient` with as little other stuff as possible for the more pressing use case? We can get that merged quick and then refactor as needed.

Thanks
</blockquote></details>

The reason I have it this way is the extremely strict way Tink makes you use their aead interface for private-key kms storage (for the code that is waiting on this in consent-signaling).   I thought passing around the Aead was more useful than the keyURI because Eli had mentioned another requirement of "What we need from US11 is just a way to encrypt/decrypt an arbitrary bytestring using a master key in a KMS"--which the Aead interface + TinkAead class provides.  It can also be reused for AeadStorageClient and the upcoming PrivateKeyStorage

What I wrote satisfies what you want above.  I figured your makeKmsStorageClient was more pseudo code for PanelMatch to write in their bringup. But here is the same code block, using my implementation with only a minor change to the return statement...

```kotlin
/** Returns a wrapper [StorageClient] that encrypts/decrypts blob data using the specified KMS [keyUri]. */
fun makeKmsStorageClient(storageClient: StorageClient, keyUri: String): StorageClient {
  val aead = KmsClients.get(keyUri).getAead(keyUri)
  return AeadStorageClient(TinkAead(aead), storageClient)
}
```

If the only sticking point is me extending the com.google.crypto.tink.Aead interface in this interface I can remove that and wrap that in a different way later (in a refactor).

efoxepstein (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/45)* status: all files reviewed, 5 unresolved discussions (waiting on @corbantek, @duliomatos1, @SanjayVas, and @wangyaopw)

___
*[src/main/kotlin/org/wfanet/measurement/common/crypto/Aead.kt, line 21 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/45#-MlLZTiB30Tyh-uWAm6Q:-MlV1OGv7meG5o9MvV_R:b-tis34j) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/12e380a9005c91310220930fece4f0228267e37f/src/main/kotlin/org/wfanet/measurement/common/crypto/Aead.kt#L21)):*
<details><summary><i>Previously, corbantek (Kyle Johnson) wrote…</i></summary><blockquote>

The reason I have it this way is the extremely strict way Tink makes you use their aead interface for private-key kms storage (for the code that is waiting on this in consent-signaling).   I thought passing around the Aead was more useful than the keyURI because Eli had mentioned another requirement of "What we need from US11 is just a way to encrypt/decrypt an arbitrary bytestring using a master key in a KMS"--which the Aead interface + TinkAead class provides.  It can also be reused for AeadStorageClient and the upcoming PrivateKeyStorage

What I wrote satisfies what you want above.  I figured your makeKmsStorageClient was more pseudo code for PanelMatch to write in their bringup. But here is the same code block, using my implementation with only a minor change to the return statement...

```kotlin
/** Returns a wrapper [StorageClient] that encrypts/decrypts blob data using the specified KMS [keyUri]. */
fun makeKmsStorageClient(storageClient: StorageClient, keyUri: String): StorageClient {
  val aead = KmsClients.get(keyUri).getAead(keyUri)
  return AeadStorageClient(TinkAead(aead), storageClient)
}
```

If the only sticking point is me extending the com.google.crypto.tink.Aead interface in this interface I can remove that and wrap that in a different way later (in a refactor).  
</blockquote></details>

@corbantek, my bad, our actual use case is an Aead-wrapped StorageClient, I was only asking for the ability to encrypt/decrypt arbitrary bytestrings because I thought it could be delivered faster.

This PR is good enough for me now. @sanjayvas, can we get this merged and then iterate more via future PRs or in a meeting?
___
*[src/main/kotlin/org/wfanet/measurement/common/crypto/TinkAead.kt, line 21 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/45#-MlV1uuX3_Ju3RiKkODx:-MlV1uuX3_Ju3RiKkODy:biel420) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/17b9798674a4b9c271b78a44f7975b0e59d0d701/src/main/kotlin/org/wfanet/measurement/common/crypto/TinkAead.kt#L21)):*
> ```Kotlin
> 
> /** Tink specific implementation of [Aead] */
> class TinkAead(val aead: TinkAeadInterface) : Aead {
> ```

`private val`?
___
*[src/main/kotlin/org/wfanet/measurement/storage/Store.kt, line 50 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/45#-MlV2OU_0JjAd-zTB4-h:-MlV2OU_0JjAd-zTB4-i:b4xmnsj) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/17b9798674a4b9c271b78a44f7975b0e59d0d701/src/main/kotlin/org/wfanet/measurement/storage/Store.kt#L50)):*
> ```Kotlin
>    * @return [Blob] with a key derived from [context]
>    */
>   open suspend fun write(context: T, content: Flow<ByteString>): Blob {
> ```

Maybe revert the the changes to this file (or move these changes to another PR)?

SanjayVas (approved):
> Reviewed 1 of 8 files at r1, 2 of 5 files at r2, 4 of 4 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/45)* status: all files reviewed, 2 unresolved discussions (waiting on @duliomatos1, @efoxepstein, and @wangyaopw)

___
*[src/main/kotlin/org/wfanet/measurement/common/crypto/Aead.kt, line 21 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/45#-MlLZTiB30Tyh-uWAm6Q:-MlVnxS5BTYGfFRPFyRz:bk1fxbm) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/12e380a9005c91310220930fece4f0228267e37f/src/main/kotlin/org/wfanet/measurement/common/crypto/Aead.kt#L21)):*
> This PR is good enough for me now. @SanjayVas, can we get this merged and then iterate more via future PRs or in a meeting?

Sure. I wasn't clear on panel match requirements before, so let's get this merged to unblock for now.

PR #48, opened 2021-10-13
Add SigningKeyStore for storing signing private keys.

SanjayVas:
Each private key is stored with its associated certificate. This ensures that the key type and signing algorithm are stored along with the private key material.

A private key read from this class is wrapped in a SigningKeyHandle. This prevents the private key material from being exposed to anything other than signing operations.

This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/48)

https://rally1.rallydev.com/#/?detail=/task/611718944581&fdp=true

### DIFF
--- a/src/main/kotlin/org/wfanet/measurement/common/crypto/BUILD.bazel
+++ b/src/main/kotlin/org/wfanet/measurement/common/crypto/BUILD.bazel
@@ -30,8 +30,29 @@ kt_jvm_library(
 )
 
 kt_jvm_library(
+    name = "signing_key_storage",
+    srcs = [
+        "SigningKeyHandle.kt",
+        "SigningKeyStore.kt",
+    ],
+    associates = [":security_provider"],
+    deps = [
+        ":signatures",
+        ":signed_blob",
+        "//imports/java/com/google/protobuf",
+        "//imports/kotlin/kotlinx/coroutines:core",
+        "//src/main/kotlin/org/wfanet/measurement/common",
+        "//src/main/kotlin/org/wfanet/measurement/storage:client",
+        "//src/main/kotlin/org/wfanet/measurement/storage:store",
+    ],
+)
+
+kt_jvm_library(
     name = "security_provider",
-    srcs = ["SecurityProvider.kt"],
+    srcs = [
+        "PemIo.kt",
+        "SecurityProvider.kt",
+    ],
     deps = [
         "//imports/java/com/google/protobuf",
         "//imports/java/org/conscrypt",
@@ -46,6 +67,17 @@ kt_jvm_library(
         "//imports/java/com/google/protobuf",
         "//src/main/kotlin/org/wfanet/measurement/common/crypto:invalid_signature_exception",
         "//src/main/kotlin/org/wfanet/measurement/common/crypto:security_provider",
+    ],
+)
+
+kt_jvm_library(
+    name = "signed_blob",
+    srcs = ["SignedBlob.kt"],
+    deps = [
+        ":signatures",
+        "//imports/java/com/google/protobuf",
+        "//imports/kotlin/kotlinx/coroutines:core",
+        "//src/main/kotlin/org/wfanet/measurement/storage:client",
     ],
 )
 
--- a/src/main/kotlin/org/wfanet/measurement/common/crypto/PemIo.kt
+++ b/src/main/kotlin/org/wfanet/measurement/common/crypto/PemIo.kt
@@ -1,0 +1,138 @@
+// Copyright 2021 The Cross-Media Measurement Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package org.wfanet.measurement.common.crypto
+
+import java.io.BufferedReader
+import java.io.ByteArrayInputStream
+import java.io.File
+import java.io.IOException
+import java.io.InputStream
+import java.io.InputStreamReader
+import java.io.OutputStream
+import java.nio.charset.StandardCharsets
+import java.security.PrivateKey
+import java.security.cert.X509Certificate
+import java.security.spec.PKCS8EncodedKeySpec
+import java.util.Base64
+import kotlin.jvm.Throws
+import org.wfanet.measurement.common.base64Decode
+
+private const val LINE_FEED: Byte = 0x0A
+private const val LINE_FEED_INT: Int = LINE_FEED.toInt()
+private const val BEGIN_PREFIX = "-----BEGIN "
+private const val END_PREFIX = "-----END "
+private const val SUFFIX = "-----"
+private const val LINE_LENGTH = 64
+private val CHARSET = StandardCharsets.US_ASCII
+private val base64Encoder = Base64.getMimeEncoder(LINE_LENGTH, byteArrayOf(LINE_FEED))
+
+private enum class PemType(private val header: String) {
+  CERTIFICATE("CERTIFICATE"),
+  PRIVATE_KEY("PRIVATE KEY");
+
+  val begin = BEGIN_PREFIX + header + SUFFIX
+  val end = END_PREFIX + header + SUFFIX
+}
+
+/**
+ * Writer for outputting objects in PEM format.
+ *
+ * See [RFC 7468](https://datatracker.ietf.org/doc/html/rfc7468)
+ */
+internal class PemWriter(private val output: OutputStream) : AutoCloseable {
+  @Throws(IOException::class)
+  private fun write(type: PemType, der: ByteArray) {
+    with(output) {
+      write(type.begin.toByteArray(CHARSET))
+      write(LINE_FEED_INT)
+      write(base64Encoder.encode(der))
+      write(LINE_FEED_INT)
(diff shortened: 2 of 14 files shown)


SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/48)* status: 17 of 22 files reviewed, 10 unresolved discussions (waiting on @efoxepstein, @jonmolle, and @SanjayVas)

___
*[src/main/kotlin/org/wfanet/measurement/common/crypto/PemIo.kt, line 29 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/48#-Mm4qBx_4sN8w-7aHWKn:-MmJj0Hp7zWE7krMur5R:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/143989433c390367bdc0390f89ff045559753cd0/src/main/kotlin/org/wfanet/measurement/common/crypto/PemIo.kt#L29)):*
<details><summary><i>Previously, efoxepstein (Eli Fox-Epstein) wrote…</i></summary><blockquote>

Somewhere in this file can you link to the PEM spec?
</blockquote></details>

Done.
___
*[src/main/kotlin/org/wfanet/measurement/common/crypto/PemIo.kt, line 45 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/48#-Mm4oOSQ8EuRRbqq-8nf:-MmJj15bCvhGa4T8Sbow:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/143989433c390367bdc0390f89ff045559753cd0/src/main/kotlin/org/wfanet/measurement/common/crypto/PemIo.kt#L45)):*
<details><summary><i>Previously, efoxepstein (Eli Fox-Epstein) wrote…</i></summary><blockquote>

KDoc please
</blockquote></details>

Done.
___
*[src/main/kotlin/org/wfanet/measurement/common/crypto/PemIo.kt, line 45 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/48#-Mm4pFTZCpN0SEwym4HV:-MmJj6a10Ycm_ldDUsZW:b-bzfyjq) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/143989433c390367bdc0390f89ff045559753cd0/src/main/kotlin/org/wfanet/measurement/common/crypto/PemIo.kt#L45)):*
<details><summary><i>Previously, efoxepstein (Eli Fox-Epstein) wrote…</i></summary><blockquote>

Question out of curiosity because I don't know the right answer: when is it better to wrap an OutputStream like this and when is it better to provide a `Externalizable` wrapper for X509Certificate and PrivateKey? Or what about just `fun OutputStream.write(x05Certificate: X509Certificate)`?
</blockquote></details>

There are multiple standards for serializing these objects. e.g. we use DER encoding for certificates in the API. The main reason to use PEM here is that's it's a common way to store a certificate and its private key together (there are OpenSSL commands for example that use this format). Another would be PKCS#12, but that's unnecessarily complicated.

The above interfaces imply having a single canonical serialization format, or at the very least having that format determined by the OutputStream. Also I think that the types used by Externalizable (e.g. `ObjectInputStream`) expect a specific header.
___
*[src/main/kotlin/org/wfanet/measurement/common/crypto/PemIo.kt, line 74 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/48#-Mm4qQ6b4iPaNGZRvtq7:-MmJsUIx20DBx0mg9k0M:b-i9ixc0) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/143989433c390367bdc0390f89ff045559753cd0/src/main/kotlin/org/wfanet/measurement/common/crypto/PemIo.kt#L74)):*
<details><summary><i>Previously, efoxepstein (Eli Fox-Epstein) wrote…</i></summary><blockquote>

It feels a little asymmetric that this operates on a Sequence instead of an InputStream. (or that PemWriter doesn't just produce a Sequence<String>)
</blockquote></details>

Done. Used streams for both.

I was just trying to use more idiomatic Kotlin with using a Sequence to wrap potentially blocking reads, but the blocking nature means that you really want to tie an output sequence to an actual stream.
___
*[src/main/kotlin/org/wfanet/measurement/common/crypto/PemIo.kt, line 111 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/48#-Mm4r8Hd4hTPUJjqg412:-MmKDX427fV5eRJWp-Qf:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/143989433c390367bdc0390f89ff045559753cd0/src/main/kotlin/org/wfanet/measurement/common/crypto/PemIo.kt#L111)):*
<details><summary><i>Previously, efoxepstein (Eli Fox-Epstein) wrote…</i></summary><blockquote>

This feels too clever, I'd prefer just a simple while-loop. (Also, we don't want to stackoverflow because a file starts with 10000 empty lines)
</blockquote></details>

Done.
___
*[src/main/kotlin/org/wfanet/measurement/common/crypto/Signatures.kt, line 58 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/48#-Mm4rlR-1_DKvxZ0L51U:-MmKN5iM8nxpgtbUBpV8:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/143989433c390367bdc0390f89ff045559753cd0/src/main/kotlin/org/wfanet/measurement/common/crypto/Signatures.kt#L58)):*
<details><summary><i>Previously, efoxepstein (Eli Fox-Epstein) wrote…</i></summary><blockquote>

This also feels too clever, is it any different than this (which I find personally much easier to read):

```
val signer = newSigner()
collect { bytes ->
  action(bytes)
  signer.update(bytes.asReadOnlyByteBuffer())
}
return signer.sign().toByteString()
```
</blockquote></details>

Done.
___
*[src/main/kotlin/org/wfanet/measurement/common/crypto/Signatures.kt, line 112 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/48#-Mm4sVeC-NjNW1AUE4L2:-MmKN6NJ0m6vfmTe1bWs:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/143989433c390367bdc0390f89ff045559753cd0/src/main/kotlin/org/wfanet/measurement/common/crypto/Signatures.kt#L112)):*
<details><summary><i>Previously, efoxepstein (Eli Fox-Epstein) wrote…</i></summary><blockquote>

Ditto
</blockquote></details>

Done.
___
*[src/main/kotlin/org/wfanet/measurement/common/crypto/SignedBlob.kt, line 59 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/48#-Mm4sxjhEFOVa8EyFuv6:-MmKOhqt3TrtX8z5Dlde:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/143989433c390367bdc0390f89ff045559753cd0/src/main/kotlin/org/wfanet/measurement/common/crypto/SignedBlob.kt#L59)):*
<details><summary><i>Previously, efoxepstein (Eli Fox-Epstein) wrote…</i></summary><blockquote>

Why is `apply` necessary? Isn't the lambda's return type `Unit`?
</blockquote></details>

Done.
___
*[src/main/kotlin/org/wfanet/measurement/common/crypto/SignedBlob.kt, line 61 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/48#-Mm4t5Mk3Ks7wxM4wTaJ:-MmKNAZV9b08HS68LtjW:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/143989433c390367bdc0390f89ff045559753cd0/src/main/kotlin/org/wfanet/measurement/common/crypto/SignedBlob.kt#L61)):*
<details><summary><i>Previously, efoxepstein (Eli Fox-Epstein) wrote…</i></summary><blockquote>

Do we have a `.toByteString()` for whatever `sign` returns?
</blockquote></details>

Done.
___
*[src/main/kotlin/org/wfanet/measurement/common/crypto/SigningKeyHandle.kt, line 27 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/48#-Mm4tTXQBrLXwkGtSi4p:-MmKNisE7bM59eQGpBTq:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/143989433c390367bdc0390f89ff045559753cd0/src/main/kotlin/org/wfanet/measurement/common/crypto/SigningKeyHandle.kt#L27)):*
<details><summary><i>Previously, efoxepstein (Eli Fox-Epstein) wrote…</i></summary><blockquote>

Maybe just `return privateKey.sign(data)`? I think we have a helper for that.
</blockquote></details>

Done.

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/48)* status: all files reviewed, 3 unresolved discussions (waiting on @corbantek, @jonmolle, and @stevenwarejones)

___
*[src/main/kotlin/org/wfanet/measurement/common/crypto/Signatures.kt, line 137 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/48#-MmTphfh3zV_hpfn2hz8:-MmU-gUyA_grw0nAM1YJ:bqbd6j3) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/d897755092908f95fe5843fad2fdbe3024d25cd4/src/main/kotlin/org/wfanet/measurement/common/crypto/Signatures.kt#L137)):*
<details><summary><i>Previously, efoxepstein (Eli Fox-Epstein) wrote…</i></summary><blockquote>

We might actually want to replace it or at least return some compound exception -- an invalid signature seems like something that should never be suppressed.

Is there something like `throw OhNoMultipleExceptionsHappened(e, InvalidSignatureException("Signature is invalid"))`?
</blockquote></details>

Another exception means that the flow did not complete successfully. Therefore it doesn't make sense to even attempt to verify the signature.

PR #50, opened 2021-10-13
Prevent 0 from being generated.

tristanvuong2021:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/50)


tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/50)* status: 1 of 3 files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[src/main/kotlin/org/wfanet/measurement/common/identity/Identifiers.kt, line 55 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/50#-Mm3uE9t-MOgOoVfkDAq:-MmK61IG7P2GpRfBCob_:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/ff819f3a24f862d8ae15779d95e422c4e8051b94/src/main/kotlin/org/wfanet/measurement/common/identity/Identifiers.kt#L55)):*

Should have the same check here.
</blockquote></details>

Done.

PR #54, opened 2021-10-29
Don't add slash in the beginning of the blob key.

wangyaopw:
The blob key works like a resource name, which has no leading slash in
the beginning.

---
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/54)


SanjayVas (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/54)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @stevenwarejones)

PR #83, opened 2022-01-19
Extend ServerCall.Listener to enable usage of rpc calls in ServerInterceptor without blocking

tristanvuong2021:
…erceptor without blocking.

---
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/83)


SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/83)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @stevenwarejones and @tristanvuong2021)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/83#-MtsdnLX6Iz1H-hYYxz-:-MtsdnLY1GrRIrhr_s50:b-557292) (no related file):*
Can you describe exactly what was happening here and how this helps fix it? I'm not entirely certain how having the delegate initialized late vs. in the constructor addresses the issue.

---

tristanvuong2021:
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/83#-MtsdnLX6Iz1H-hYYxz-:-MtsfSBX8SPU0FKd9uly:b-58ochn) (no related file):*

Can you describe exactly what was happening here and how this helps fix it? I'm not entirely certain how having the delegate initialized late vs. in the constructor addresses the issue.
</blockquote></details>

@sanjayvas I need to get rid of runblocking, but since rpc calls are suspend functions they need to be run in a coroutine. If I just create a coroutine and use the default listener, since I am not blocking the interceptor will move on regardless of whether or not the rpc call finishes and the following code dependent on it runs. With this listener, the interceptor cannot move on to the next step until the rpc call finishes. It is best to just look at the way I use it in the other pull request I have.

---

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/83)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @stevenwarejones and @tristanvuong2021)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/83#-MtsdnLX6Iz1H-hYYxz-:-MttlrrUBhPOfkaH7G1y:b-uqp6ue) (no related file):*
<details><summary><i>Previously, tristanvuong2021 (Tristan Vuong) wrote…</i></summary><blockquote>

@sanjayvas I need to get rid of runblocking, but since rpc calls are suspend functions they need to be run in a coroutine. If I just create a coroutine and use the default listener, since I am not blocking the interceptor will move on regardless of whether or not the rpc call finishes and the following code dependent on it runs. With this listener, the interceptor cannot move on to the next step until the rpc call finishes. It is best to just look at the way I use it in the other pull request I have.
</blockquote></details>

I see how it's being used, but I can't verify how it solves the problem and if this is the best approach without knowing what caused the deadlock. e.g. which two things were interfering?

Given that the deadlock was appearing in GHA and not on your local machine, my guess was that it had something to do with [`Dispatchers.Default`](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-dispatchers/-default.html) and the number of cores. This implies that two things that conflict are using the same dispatcher, e.g. if interceptors run in the same dispatcher as that used to handle incoming RPCs or having the public server interceptor and internal server interceptor both blocked. I figured this could be solved simply by ensuring all RPCs made by all interceptors use Dispatchers.IO.

---

tristanvuong2021:
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/83#-MtsdnLX6Iz1H-hYYxz-:-Mttq4iD-VA13ZLsclyU:bwlcdpo) (no related file):*

I see how it's being used, but I can't verify how it solves the problem and if this is the best approach without knowing what caused the deadlock. e.g. which two things were interfering?

Given that the deadlock was appearing in GHA and not on your local machine, my guess was that it had something to do with [`Dispatchers.Default`](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-dispatchers/-default.html) and the number of cores. This implies that two things that conflict are using the same dispatcher, e.g. if interceptors run in the same dispatcher as that used to handle incoming RPCs or having the public server interceptor and internal server interceptor both blocked. I figured this could be solved simply by ensuring all RPCs made by all interceptors use Dispatchers.IO.
</blockquote></details>

Runblocking always blocks regardless of which dispatcher is used: https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/run-blocking.html

---

tristanvuong2021:
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/83#-MtsdnLX6Iz1H-hYYxz-:-MttqgUs5IZcJn3UQFP2:b-2q4z4h) (no related file):*
<details><summary><i>Previously, tristanvuong2021 (Tristan Vuong) wrote…</i></summary><blockquote>

Runblocking always blocks regardless of which dispatcher is used: https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/run-blocking.html
</blockquote></details>

Changing the dispatcher just changes what runs the code inside the runblocking block, but doesn't prevent the runblocking itself trying to block the current thread.

---

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/83)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @stevenwarejones and @tristanvuong2021)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/83#-MtsdnLX6Iz1H-hYYxz-:-MttwmOf6oejkL7nHAzV:bsm0x1d) (no related file):*
<details><summary><i>Previously, tristanvuong2021 (Tristan Vuong) wrote…</i></summary><blockquote>

Changing the dispatcher just changes what runs the code inside the runblocking block, but doesn't prevent the runblocking itself trying to block the current thread.
</blockquote></details>

I understood that about `runBlocking`. It looks like what I was missing was that you must not block in `interceptCall`, so this is deferring the blocking until after `interceptCall` returns.

---
___
*[src/main/kotlin/org/wfanet/measurement/common/grpc/DeferredListener.kt, line 20 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/83#-MttyL6kCBMG4eCASPf1:-MttyL6kCBMG4eCASPf2:b-4ovh9j) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/0ba2e245fb8fc827c6765fc35f77bc778209883e/src/main/kotlin/org/wfanet/measurement/common/grpc/DeferredListener.kt#L20)):*
> ```Kotlin
> 
> /** gRPC [ServerCall.Listener] that defers events */
> class DeferredListener<ReqT> : ServerCall.Listener<ReqT>() {
> ```

Maybe `DeferredForwardingListener`? That key behavior is also missing from the description: that this defers forwarding events to a delegate until `setDelegate` is called.



_Code quote:_
```Kotlin
DeferredListener
```

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/83)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @SanjayVas and @stevenwarejones)

___
*[src/main/kotlin/org/wfanet/measurement/common/grpc/DeferredListener.kt, line 20 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/83#-MttyL6kCBMG4eCASPf1:-Mtu2C0G0c2pWT9OrrLi:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/0ba2e245fb8fc827c6765fc35f77bc778209883e/src/main/kotlin/org/wfanet/measurement/common/grpc/DeferredListener.kt#L20)):*

Maybe `DeferredForwardingListener`? That key behavior is also missing from the description: that this defers forwarding events to a delegate until `setDelegate` is called.
</blockquote></details>

Done.

SanjayVas (approved):
> Reviewed 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/83)* status: all files reviewed, 1 unresolved discussion (waiting on @stevenwarejones and @tristanvuong2021)

___
*[src/main/kotlin/org/wfanet/measurement/common/grpc/DeferredForwardingListener.kt, line 20 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/83#-MtuBVmv6IMQRcUT_Rhz:-MtuBVmv6IMQRcUT_Ri-:bz0hvm3) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/c045bb8cf828623531239297deb62bf316f57b97/src/main/kotlin/org/wfanet/measurement/common/grpc/DeferredForwardingListener.kt#L20)):*
> ```Kotlin
> 
> /**
>  * gRPC [ServerCall.Listener] that defers forwarding events to a delegate until setDelegate is
> ```

nit: Use KDoc reference



_Code quote:_
```Kotlin
setDelegate
```

stevenwarejones (approved):
> Reviewed 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/83)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @tristanvuong2021)

PR #86, opened 2022-01-25
randomize storage keys for abstract storage client tests

stevenwarejones:
Some of these tests already using randomized keys. This extends this to the rest of the tests for build environments that run these tests multiple times in parallel.

---
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/86)

### DIFF
--- a/src/main/kotlin/org/wfanet/measurement/storage/testing/AbstractStorageClientTest.kt
+++ b/src/main/kotlin/org/wfanet/measurement/storage/testing/AbstractStorageClientTest.kt
@@ -40,7 +40,7 @@ abstract class AbstractStorageClientTest<T : StorageClient> {
 
   @Test
   fun `createBlob returns new readable blob`() = runBlocking {
-    val blobKey = "new-blob"
+    val blobKey = "new-blob-" + random.nextInt().toString()
 
     val blob = storageClient.createBlob(blobKey, testBlobContent)
 
@@ -49,13 +49,13 @@ abstract class AbstractStorageClientTest<T : StorageClient> {
 
   @Test
   fun `getBlob returns null for non-existant blob`() {
-    val blobKey = "non-existant-blob"
+    val blobKey = "non-existant-blob-" + random.nextInt().toString()
     assertThat(storageClient.getBlob(blobKey)).isNull()
   }
 
   @Test
   fun `getBlob returns readable Blob`() = runBlocking {
-    val blobKey = "blob-to-get"
+    val blobKey = "blob-to-get-" + random.nextInt().toString()
     storageClient.createBlob(blobKey, testBlobContent)
 
     val blob = assertNotNull(storageClient.getBlob(blobKey))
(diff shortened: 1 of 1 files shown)


SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/86)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @SanjayVas and @stevenwarejones)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/86#-MuHaAYv7PonVYWsF1rk:-MuHaAYv7PonVYWsF1rl:b-asqgui) (no related file):*
> This extends this to the rest of the tests for build environments that run these tests multiple times in parallel.

Can you clarify why this would be needed? Our tests should already be isolated such that they can be run in parallel. Is there some implementation of this that is e.g. sharing the underlying storage across runs?

---

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/86)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @SanjayVas and @stevenwarejones)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/86#-MuHaAYv7PonVYWsF1rk:-MuIGDV9CFLm_e4R84L3:b-ovael3) (no related file):*
My basic point is that the implementing test class should be guaranteeing isolation. We shouldn't need to randomize blob names for this purpose.

> The filesystem test already in the repo can't be run 2x in parallel either.

It can't? It uses the TemporaryDirectory rule, so it should be creating a separate temp dir for each run.

---

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/86)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/86#-MuHaAYv7PonVYWsF1rk:-MuO4pYuES6-qkzVFuz8:b-4xrds1) (no related file):*

My basic point is that the implementing test class should be guaranteeing isolation. We shouldn't need to randomize blob names for this purpose.

> The filesystem test already in the repo can't be run 2x in parallel either.

It can't? It uses the TemporaryDirectory rule, so it should be creating a separate temp dir for each run.
</blockquote></details>

You're 100% right. Not sure what I was thinking... too much multi-tasking.... anyways, I do have a use case with storage that fails with the current tests. If you think its bad practice to encourage it, we can leave it as is, but I'd prefer to merge this.

---

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/86)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @SanjayVas and @stevenwarejones)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/86#-MuHaAYv7PonVYWsF1rk:-MuRySsJ65w38HPmsQOD:b-b0v95o) (no related file):*

You're 100% right. Not sure what I was thinking... too much multi-tasking.... anyways, I do have a use case with storage that fails with the current tests. If you think its bad practice to encourage it, we can leave it as is, but I'd prefer to merge this.
</blockquote></details>

I think the priority would be figuring out how to fix that test, as the test setup itself should ensure isolation. That is to say that this test should be free to use the same keys across calls with the assumption that test runs cannot conflict.

---

Issue #51, opened 2021-10-18
Drop bufferSizeBytes parameter from StorageClient.Blob.read.


SanjayVas:
This parameter is not actually useful, as it's easy for the caller to rebuffer after the fact (change the Flow value size). Each `StorageClient` implementation should determine its own optimal buffer sizing for reads/writes to the underlying storage system.

PR #92, opened 2022-02-08
Drop StorageClient.defaultBufferSizeBytes.

SanjayVas:
Fixes world-federation-of-advertisers/common-jvm#51

---
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/92)


stevenwarejones (changes requested):
> Reviewed 22 of 22 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/92)* status: all files reviewed, 4 unresolved discussions (waiting on @SanjayVas)

___
*[src/main/kotlin/org/wfanet/measurement/common/Bytes.kt, line 146 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/92#-MvQ0JrW5aTOIyfdHnhd:-MvQ0JrW5aTOIyfdHnhe:b2lvic8) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/da021ad82dc2bd5f58ada10da324f5aa10219564/src/main/kotlin/org/wfanet/measurement/common/Bytes.kt#L146)):*
> ```Kotlin
>  */
> fun ByteBuffer.asBufferedFlow(bufferSize: Int): Flow<ByteString> {
>   require(bufferSize > 0)
> ```

the following 8 lines are duplicated a couple times. do you want to combine into a helper?
___
*[src/main/kotlin/org/wfanet/measurement/storage/StorageClient.kt, line 30 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/92#-MvQ19tN5kpVl_mOQzqS:-MvQ19tN5kpVl_mOQzqT:bc0rqwd) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/c91060e7f90078e9cba118452baee9440a9d944f/src/main/kotlin/org/wfanet/measurement/storage/StorageClient.kt#L30)):*
> ```Kotlin
> interface StorageClient {
>   /** Default size in bytes of each [Flow] value. */
>   val defaultBufferSizeBytes: Int
> ```

I'm not sure I like this. Isn't it better to force StorageClient to specify read and write buffer sizes?
___
*[src/main/kotlin/org/wfanet/measurement/storage/StorageClient.kt, line 58 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/92#-MvQ0iga-ppbwUWIiUju:-MvQ0iga-ppbwUWIiUjv:bco6y5k) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/c91060e7f90078e9cba118452baee9440a9d944f/src/main/kotlin/org/wfanet/measurement/storage/StorageClient.kt#L58)):*
> ```Kotlin
>  * [StorageClient.defaultBufferSizeBytes] size.
>  */
> suspend fun StorageClient.writeBlob(blobKey: String, content: ByteString) =
> ```

i like these convenience functions
___
*[src/main/kotlin/org/wfanet/measurement/storage/testing/AbstractStorageClientTest.kt, line 70 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/92#-MvQ0xpw87bGxR3aIlO_:-MvQ0xpw87bGxR3aIlOa:b-dwluu) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/da021ad82dc2bd5f58ada10da324f5aa10219564/src/main/kotlin/org/wfanet/measurement/storage/testing/AbstractStorageClientTest.kt#L70)):*
> ```Kotlin
>     val blobKey = "blob-to-check-size-" + random.nextInt().toString()
> 
>     val blob = storageClient.writeBlob(blobKey, flowOf(testBlobContent))
> ```

if you keep the convenience function, these can stay

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/92)* status: 17 of 22 files reviewed, 3 unresolved discussions (waiting on @stevenwarejones)

___
*[src/main/kotlin/org/wfanet/measurement/common/Bytes.kt, line 146 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/92#-MvQ0JrW5aTOIyfdHnhd:-MvQP8t524h_dUurrQ1f:b-s6sz2m) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/da021ad82dc2bd5f58ada10da324f5aa10219564/src/main/kotlin/org/wfanet/measurement/common/Bytes.kt#L146)):*

the following 8 lines are duplicated a couple times. do you want to combine into a helper?
</blockquote></details>

I believe it's only twice, and each uses different members for its conditions.
___
*[src/main/kotlin/org/wfanet/measurement/storage/StorageClient.kt, line 30 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/92#-MvQ19tN5kpVl_mOQzqS:-MvQPuoE-KQR9TzLPmlE:bvf1yyl) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/c91060e7f90078e9cba118452baee9440a9d944f/src/main/kotlin/org/wfanet/measurement/storage/StorageClient.kt#L30)):*

I'm not sure I like this. Isn't it better to force StorageClient to specify read and write buffer sizes?
</blockquote></details>

It turns out the desired behavior is different for each backend. e.g. the GCS client library uses internal buffering with its own chosen optimal size for writes, so there's no point specifying a write buffer size. The async S3 client (which we should switch to) similarly has its own internal optimal buffering for reads.

It's up to each implementation to decide if/how it wants to buffer.
___
*[src/main/kotlin/org/wfanet/measurement/storage/StorageClient.kt, line 58 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/92#-MvQ0iga-ppbwUWIiUju:-MvQQkhGC29oszRY9vAq:b-4wjeln) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/c91060e7f90078e9cba118452baee9440a9d944f/src/main/kotlin/org/wfanet/measurement/storage/StorageClient.kt#L58)):*

i like these convenience functions
</blockquote></details>

Added writeBlob w/ByteString as a convenience overload.

The `read` extension signature now just matches the actual method.
___
*[src/main/kotlin/org/wfanet/measurement/storage/testing/AbstractStorageClientTest.kt, line 70 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/92#-MvQ0xpw87bGxR3aIlO_:-MvQZgTm4uSB4smpjozp:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/da021ad82dc2bd5f58ada10da324f5aa10219564/src/main/kotlin/org/wfanet/measurement/storage/testing/AbstractStorageClientTest.kt#L70)):*

if you keep the convenience function, these can stay 
</blockquote></details>

Done.

stevenwarejones (approved):
> Reviewed 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/92)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @SanjayVas)

PR #99, opened 2022-03-10
Add support for an optional prefix path in GcsStorageClient

jonmolle:
This is used to avoid collision when using a static bucket for multiple days' of panel exchange shared storage. Should be a no-op for all current usages. A followup PR in panel-exchange-client will make use of this.

---
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/99)

### DIFF
--- a/src/main/kotlin/org/wfanet/measurement/gcloud/gcs/GcsStorageClient.kt
+++ b/src/main/kotlin/org/wfanet/measurement/gcloud/gcs/GcsStorageClient.kt
@@ -37,14 +37,25 @@ import org.wfanet.measurement.storage.StorageClient
 private const val BYTE_BUFFER_SIZE = BYTES_PER_MIB * 1
 
 /** Google Cloud Storage (GCS) implementation of [StorageClient] for a single bucket. */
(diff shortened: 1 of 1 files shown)


SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/99)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @jonmolle, @stevenwarejones, and @yunyeng)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/99#-MxpNARP2Ta9q-8Mgkix:-MxpNARQ-JCobg73abf9:btrwxge) (no related file):*
Why have this be specific to GcsStorageClient? Also, panel-exchange-client already has a `PrefixedStorageClient` class which wraps a `StorageClient` to add a prefix to the blob key.

---

jonmolle:
SanjayVas pointed me to an existing wrapper that does what we could use, so this PR is unnecessary. Closing.

PR #102, opened 2022-04-12
Parse SKID and AKID according to DER spec.

SanjayVas:
This includes a slight behavioral change to the subjectKeyIdentifier and authorityKeyIdentifier properties, where they will now throw an exception rather than returning null if parsing fails.

---
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/102)

### DIFF
--- a/src/main/kotlin/org/wfanet/measurement/common/Bytes.kt
+++ b/src/main/kotlin/org/wfanet/measurement/common/Bytes.kt
@@ -296,6 +296,10 @@ fun File.readByteString(): ByteString {
   return inputStream().use { input -> ByteString.readFrom(input) }
 }
 
+fun Byte.toStringHex(): String {
+  return "%2x".format(this)
+}
+
 /** Converts a hex string to its equivalent [ByteString]. */
 @Deprecated(
   "Use HexString for stronger typing",
--- a/src/main/kotlin/org/wfanet/measurement/common/crypto/BUILD.bazel
+++ b/src/main/kotlin/org/wfanet/measurement/common/crypto/BUILD.bazel
@@ -65,11 +65,13 @@ kt_jvm_library(
 kt_jvm_library(
     name = "security_provider",
     srcs = [
+        "DerEncoding.kt",
         "SecurityProvider.kt",
     ],
     deps = [
         "//imports/java/com/google/protobuf",
         "//imports/java/org/conscrypt",
+        "//src/main/kotlin/org/wfanet/measurement/common",
     ],
 )
 
--- a/src/main/kotlin/org/wfanet/measurement/common/crypto/DerEncoding.kt
+++ b/src/main/kotlin/org/wfanet/measurement/common/crypto/DerEncoding.kt
@@ -1,0 +1,179 @@
+/*
+ * Copyright 2022 The Cross-Media Measurement Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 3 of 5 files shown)


stevenwarejones (changes requested):
can you add a test that throws an exception?

> Reviewed 2 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/102#-:-N0fsdCt0gc3kfGb9iIC:b-85ty9d)* status: 2 of 5 files reviewed, 1 unresolved discussion (waiting on @SanjayVas and @stevenwarejones)

___
*[`src/test/kotlin/org/wfanet/measurement/common/crypto/SecurityProviderTest.kt` line 55 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/102#-N0fs7os7lbcHDossqjx:-N0fs7os7lbcHDossqjy:bar9p9y) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/91b8802bb7f6a1db6cedcc3d623771c12dc136bf/src/test/kotlin/org/wfanet/measurement/common/crypto/SecurityProviderTest.kt#L55)):*
> ```Kotlin
>     0xA8
>   )
> private val CLIENT_SKID =
> ```

i realize this is unused but shouldn't we be testing that it correctly reads the skid from the meta certificate as well?

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/102)* status: 2 of 5 files reviewed, 1 unresolved discussion (waiting on @stevenwarejones)

___
*[`src/test/kotlin/org/wfanet/measurement/common/crypto/SecurityProviderTest.kt` line 55 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/102#-N0fs7os7lbcHDossqjx:-N0g1m9T8sEYwLlHao5P:b8rmj0o) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/91b8802bb7f6a1db6cedcc3d623771c12dc136bf/src/test/kotlin/org/wfanet/measurement/common/crypto/SecurityProviderTest.kt#L55)):*

i realize this is unused but shouldn't we be testing that it correctly reads the skid from the meta certificate as well?
</blockquote></details>

The SKID encoding doesn't actually vary across certificates since it's just a single ASN.1 OCTET STRING, so it's encoded exactly the same way in the Meta cert as in the test cert generated from OpenSSL.

It turns out the bug we had in the previous implementation was actually just that a [documented assumption](https://github.com/world-federation-of-advertisers/common-jvm/blob/2172e9689ddf605a0b2f424e208d04eeff989622/src/main/kotlin/org/wfanet/measurement/common/crypto/SecurityProvider.kt#L140) didn't hold, namely that the length of the extension would always be <=127 bytes. I was being lazy and didn't want to write out the more complex code for handling longer lengths if it was a case we'd never hit in practice.

stevenwarejones (approved):
You could just take a file and malform it.

> Reviewed 3 of 5 files at r1.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/102#-:-N0ge0b_CNzf0HWettl0:bkys2u3)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @SanjayVas)

SanjayVas (commented):
Right, but I don't know how to malform just the extension bytes short of using a hex editor and searching for a known pattern. Most tools for writing X.509 certificates won't let you make a malformed one.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/102#-:-N0gqrEoC3qg_m8_GJP8:b-fnlyn6)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @SanjayVas)

PR #106, opened 2022-04-19
Add rules for K8s Kustomization.

SanjayVas:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/106)


SanjayVas:
The canonical version of k8s rules currently exists in the cross-media-measurement repo.

PR #110, opened 2022-05-04
Update Bazel version to 5.1.1

iverson52000:
Update Bazel version to 5.1.1 and fix breaking changes from transitive dependencies
image

---
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/110)

PR #112, opened 2022-05-09
improve Bytes.kt compatibilty for running from java 8 but compiling from java 11

stevenwarejones:
See https://www.morling.dev/blog/bytebuffer-and-the-dreaded-nosuchmethoderror/

---
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/112)

### DIFF
--- a/src/main/kotlin/org/wfanet/measurement/common/Bytes.kt
+++ b/src/main/kotlin/org/wfanet/measurement/common/Bytes.kt
@@ -18,6 +18,7 @@ import com.google.protobuf.ByteString
 import com.google.protobuf.kotlin.toByteString
 import java.io.File
 import java.io.InputStream
+import java.nio.Buffer
 import java.nio.ByteBuffer
 import java.nio.ByteOrder
 import java.nio.channels.ReadableByteChannel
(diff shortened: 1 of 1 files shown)


SanjayVas (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/112)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @SanjayVas)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/112#-N1drlzI1AeoiFTRCPRi:-N1eNZYt8eKVOjwsOCr9:b-5p0wws) (no related file):*

Where do we need Java 8 compatibility? While we currently use language version 8 (mainly because it doesn't matter much when using Kotlin), but I believe we've expected at least JDK 11 at runtime which IIUC doesn't have this issue.

If we do need Java 8 compatibility, then we should see about what we need for Bazel flags to use the appropriate `--release` option for javac as described in the article. We should also document explicitly that we intend to be compatible with Java 8 in this case. I know that there are some changes to how Bazel builds Java code in Bazel 5 if that has any impact. See https://github.com/world-federation-of-advertisers/common-jvm/pull/110
</blockquote></details>

Looks like `--release` isn't well-supported due to also blocking access to internal APIs like `sun.misc.Unsafe`, which is depended upon by protobuf.

---

PR #113, opened 2022-05-09
change file system storage client to use InputStream.asFlow rather than ReadableByteChannel.asFlow

stevenwarejones:
…hannel

---
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/113)


SanjayVas (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/113)* status: all files reviewed, 1 unresolved discussion (waiting on @stevenwarejones)

___
*[`-- commits` line 3 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/113#-N1ikhJP6-4e5YXfWpSv):*
Can you make this more clear in the description that you're using InputStream.asFlow rather than ReadableByteChannel.asFlow?

It's technically not a question of whether you have to convert it, but rather which implementation you choose to use. `FileInputStream` in particular has a `getChannel` method to give you a `ReadableByteChannel`. I'm not actually certain whether either implementation is "better" when it comes to e.g. read performance, so swapping between them is fine.



_Code quote:_
```
file system storage client does not need to convert input stream to channel
```

PR #120, opened 2022-05-24
move jni components to comon-jvm from panel-exchange-client

stevenwarejones:


robinsons (approved):
> Reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/120)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @SanjayVas)

SanjayVas (changes requested):
> Reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/120)* status: all files reviewed, 1 unresolved discussion (waiting on @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/common/LoadLibrary.kt` line 38 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/120#-N2sK2d-8bMOIBJNmCBz:-N2sK2d05X-guXE7b2Lr:b-laao32) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/f21052883cfd13f6d4a014b94afdbefa085bdb10/src/main/kotlin/org/wfanet/measurement/common/LoadLibrary.kt#L38)):*
> ```Kotlin
>     Paths.get(
>         System.getProperty("java.io.tmpdir"),
>         "org-wfanet-panelmatch-common-beam-LoadLibrary",
> ```

This seems panelmatch-specific.

If we want something in common-jvm, it should load libraries from some "natural" location within JAR resources, e.g. wherever Bazel puts native deps of a `java_library`. This change should coincide with updating rules_swig to put native deps in this location.

The `loadLibrary` function in https://github.com/world-federation-of-advertisers/common-jvm/blob/main/src/main/kotlin/org/wfanet/measurement/common/Runfiles.kt handles loading libraries based on the current implementation of rules_swig.



_Code quote:_
```Kotlin
org-wfanet-panelmatch-common-beam-LoadLibrary
```

PR #130, opened 2022-06-16
Add `get` operator extension to Spanner's `StructReader`.

SanjayVas:
This uses reified types such that the Kotlin compiler may be able to infer the correct type.


SanjayVas:
This came from what I learned writing #127.

SanjayVas:
Dropping this, as it's significantly less performant by an order of magnitude.

PR #160, opened 2022-08-24
Allow CoroutineContexts and Executors to be specified/configured.

SanjayVas:
This includes using a fixed thread pool Executor for gRPC servers. It allows more control over thread usage.

- - -
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/160)

### DIFF
--- a/build/common_jvm_maven.bzl
+++ b/build/common_jvm_maven.bzl
@@ -75,6 +75,7 @@ def common_jvm_maven_artifacts_dict():
         "info.picocli:picocli": "4.4.0",
         "junit:junit": "4.13",
         "org.conscrypt:conscrypt-openjdk-uber": "2.5.2",
+        "org.jetbrains:annotations": "23.0.0",
         "org.mockito.kotlin:mockito-kotlin": "3.2.0",
         "software.amazon.awssdk:s3": "2.17.258",
 
--- a/build/rules_kotlin/deps.bzl
+++ b/build/rules_kotlin/deps.bzl
@@ -19,11 +19,11 @@ load(
     "kotlin_repositories",
     "kotlinc_version",
 )
-load(
-    "//build:versions.bzl",
-    "JETBRAINS_ANNOTATIONS_VERSION",
-    "KOTLIN_RELEASE_VERSION",
-)
+load("//build:versions.bzl", "KOTLIN_RELEASE_VERSION")
+
+# Version of org.jetbrains:annotations that comes bundled with
+# KOTLIN_RELEASE_VERSION.
+JETBRAINS_ANNOTATIONS_VERSION = "13.0"
 
 def rules_kotlin_deps():
     compiler_release = kotlinc_version(
--- a/build/rules_kotlin/repo.bzl
+++ b/build/rules_kotlin/repo.bzl
@@ -19,7 +19,7 @@ See https://github.com/bazelbuild/rules_kotlin
 
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 load("@bazel_tools//tools/build_defs/repo:utils.bzl", "maybe")
-load("//build:versions.bzl", "JETBRAINS_ANNOTATIONS_VERSION", "KOTLIN_RELEASE_VERSION")
+load("//build:versions.bzl", "KOTLIN_RELEASE_VERSION")
 
 _JETBRAINS_KOTLIN_OVERRIDE_TARGETS = {
     "org.jetbrains.kotlin:kotlin-stdlib": "@com_github_jetbrains_kotlin//:kotlin-stdlib",
@@ -32,9 +32,14 @@ _JETBRAINS_KOTLIN_OVERRIDE_TARGETS = {
 
 _JETBRAINS_OVERRIDE_TARGETS = {
     "org.jetbrains:annotations": "@com_github_jetbrains_kotlin//:annotations",
-}
+}  # @unused
 
-RULES_KOTLIN_OVERRIDE_TARGETS = dict(_JETBRAINS_KOTLIN_OVERRIDE_TARGETS.items() + _JETBRAINS_OVERRIDE_TARGETS.items())
+# Override targets for rules_kotlin.
+#
+# Despite the fact that the Kotlin compiler release bundles JetBrains,
+# annotations, we intentionally do not include it as an override target since
+# as of the 1.6 compiler release it is quite an old version (13).
+RULES_KOTLIN_OVERRIDE_TARGETS = _JETBRAINS_KOTLIN_OVERRIDE_TARGETS
 
 def _rules_kotlin_repo(version, sha256):
     maybe(
@@ -55,8 +60,4 @@ def rules_kotlin_maven_artifacts_dict():
         coordinates: KOTLIN_RELEASE_VERSION
         for coordinates in _JETBRAINS_KOTLIN_OVERRIDE_TARGETS.keys()
     }
-    artifacts_dict.update({
-        coordinates: JETBRAINS_ANNOTATIONS_VERSION
-        for coordinates in _JETBRAINS_OVERRIDE_TARGETS.keys()
-    })
     return artifacts_dict
--- a/build/versions.bzl
+++ b/build/versions.bzl
@@ -28,10 +28,6 @@ KOTLIN_LANGUAGE_LEVEL = "1.5"
 # See https://kotlinlang.org/docs/releases.html#release-details.
 KOTLIN_RELEASE_VERSION = "1.6.21"
 
-# Version of org.jetbrains:annotations that comes bundled with
-# KOTLIN_RELEASE_VERSION.
-JETBRAINS_ANNOTATIONS_VERSION = "13.0"
-
 # kotlinx.coroutines version.
 KOTLINX_COROUTINES_VERSION = "1.6.2"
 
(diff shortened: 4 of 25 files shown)


SanjayVas:
The linter warning can safely be ignored. It's triggering https://github.com/bazelbuild/buildtools/issues/1069.

stevenwarejones (changes requested):
> Reviewed 24 of 25 files at r1, 2 of 2 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/160)* status: all files reviewed, 3 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/tink/TinkKeyStorageProvider.kt` line 28 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/160#-NAP_MqE-Vm3JmmzDKY_:-NAP_MqE-Vm3JmmzDKYa:b-b7ueom) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/fee843f3f53c91ca94b4704b76fd043f38e4f224/src/main/kotlin/org/wfanet/measurement/common/crypto/tink/TinkKeyStorageProvider.kt#L28)):*
> ```Kotlin
> 
> class TinkKeyStorageProvider(
>   private val aeadContext: @BlockingExecutor CoroutineContext = Dispatchers.IO
> ```

what are the uses cases for needing to be able to configure aeadContext to something other than this?
___
*[`src/main/kotlin/org/wfanet/measurement/common/grpc/Channel.kt` line 39 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/160#-NAPa96922SXxIUNyzKs:-NAPa96922SXxIUNyzKt:bl3j4nk) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/fee843f3f53c91ca94b4704b76fd043f38e4f224/src/main/kotlin/org/wfanet/measurement/common/grpc/Channel.kt#L39)):*
> ```Kotlin
>   val channelBuilder =
>     NettyChannelBuilder.forTarget(target)
>       .directExecutor() // See https://github.com/grpc/grpc-kotlin/issues/263
> ```

are you noticing any performance increase by doing this?



_Code quote:_
```Kotlin
See https://github.com/grpc/grpc-kotlin/issues/263
```
___
*[`src/main/kotlin/org/wfanet/measurement/common/grpc/CommonServer.kt` line 63 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/160#-NAPaMgA-6XaGzN7TDE-:-NAPaMgA-6XaGzN7TDE0:b-x49ttw) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/fee843f3f53c91ca94b4704b76fd043f38e4f224/src/main/kotlin/org/wfanet/measurement/common/grpc/CommonServer.kt#L63)):*
> ```Kotlin
>       .apply {
>         executor(executor)
>         if (sslContext != null) {
> ```

why do you prefer this over let? Isn't let more kotlin-esque?

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/160)* status: 24 of 25 files reviewed, 3 unresolved discussions (waiting on @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/tink/TinkKeyStorageProvider.kt` line 28 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/160#-NAP_MqE-Vm3JmmzDKY_:-NAfZ1FD3Ox8eBl5bQve:b-b0r7kp) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/fee843f3f53c91ca94b4704b76fd043f38e4f224/src/main/kotlin/org/wfanet/measurement/common/crypto/tink/TinkKeyStorageProvider.kt#L28)):*

what are the uses cases for needing to be able to configure aeadContext to something other than this?
</blockquote></details>

If you want more control over the threads used by this. e.g. if you want to use a cached thread pool with different parameters, or one which uses a different thread factory.

The built-in Dispatchers aren't bad, but they don't allow you to easily control and track threads. We at Google are in fact [forbidden from using them in internal projects](https://github.com/grpc/grpc-kotlin/issues/60#issuecomment-618186497), and we would instead inject one that uses a thread pool constructed from one of our frameworks.

We're *probably* fine with `Dispatchers.Default` for CPU-bound and `Dispatchers.IO` for blocking I/O, but this leaves it open so callers can tune things.
___
*[`src/main/kotlin/org/wfanet/measurement/common/grpc/Channel.kt` line 39 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/160#-NAPa96922SXxIUNyzKs:-NAfaE3nCFLPHiaYOtW1:bryfrov) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/fee843f3f53c91ca94b4704b76fd043f38e4f224/src/main/kotlin/org/wfanet/measurement/common/grpc/Channel.kt#L39)):*

are you noticing any performance increase by doing this?
</blockquote></details>

I haven't benchmarked it, but it does make sense given how grpc-kotlin uses coroutines. As mentioned in the linked issue, it should only be a problem if you have poorly-behaved interceptors (e.g. that do blocking I/O).

The [gRPC docs](https://grpc.io/docs/guides/performance/#java) recommend not using the default executors. The defaults are meant to avoid deadlocks/starvation in the case where blocking calls are done in places they shouldn't, but incur a performance penalty in well-behaved applications. See https://github.com/grpc/grpc-java/issues/7381 for more details on the default executor decision.
___
*[`src/main/kotlin/org/wfanet/measurement/common/grpc/CommonServer.kt` line 63 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/160#-NAPaMgA-6XaGzN7TDE-:-NAfdFZnDyadacfx_tcg:bq9f2q9) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/fee843f3f53c91ca94b4704b76fd043f38e4f224/src/main/kotlin/org/wfanet/measurement/common/grpc/CommonServer.kt#L63)):*
Our internal Google Kotlin best practices guide states to avoid using `?.let {...}` where smart casts would suffice, and explicitly calls out `if (namedVariable != null) { ... }` case.

>` expression?.let { something(it) }` is a common idiom in Kotlin to execute some operations on the result of an expression if expression does not evaluate to null. However, the Kotlin standard libraries almost never use this idiom: JetBrains has reached the conclusion that it generally harms readability.

I'm happy to change since we're not following this style explicitly in this project, but the advice seems sound given that it's what Google and Jetbrains, being authors of the Kotlin language, follow.

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/160)* status: 24 of 25 files reviewed, 1 unresolved discussion (waiting on @SanjayVas and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/common/grpc/CommonServer.kt` line 63 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/160#-NAPaMgA-6XaGzN7TDE-:-NAkGD2XAQ-dPK-w8vQl:bfyc21t) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/fee843f3f53c91ca94b4704b76fd043f38e4f224/src/main/kotlin/org/wfanet/measurement/common/grpc/CommonServer.kt#L63)):*

Our internal Google Kotlin best practices guide states to avoid using `?.let {...}` where smart casts would suffice, and explicitly calls out `if (namedVariable != null) { ... }` case.

>` expression?.let { something(it) }` is a common idiom in Kotlin to execute some operations on the result of an expression if expression does not evaluate to null. However, the Kotlin standard libraries almost never use this idiom: JetBrains has reached the conclusion that it generally harms readability.

I'm happy to change since we're not following this style explicitly in this project, but the advice seems sound given that it's what Google and Jetbrains, being authors of the Kotlin language, follow.
</blockquote></details>

I'm fine either way but I `?.let` is all over the place and I've actually had a number of reviews where the reviewer asked me to change it to `?.let` on this project

SanjayVas (commented):
> Dismissed @stevenwarejones from a discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/160)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/common/grpc/CommonServer.kt` line 63 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/160#-NAPaMgA-6XaGzN7TDE-:-NAkhBzd01JtVCDoYWFI:b-49bj8t) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/fee843f3f53c91ca94b4704b76fd043f38e4f224/src/main/kotlin/org/wfanet/measurement/common/grpc/CommonServer.kt#L63)):*

I'm fine either way but I `?.let` is all over the place and I've actually had a number of reviews where the reviewer asked me to change it to `?.let` on this project
</blockquote></details>

There is one case where I think `?.let` may be more readable, and that's when chaining.

PR #161, opened 2022-08-25
Add finally block to withRetriesOnEach

renjiezh:


matthewclegg (approved):
> Reviewed 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/161)* status: all files reviewed, 1 unresolved discussion (waiting on @renjiezh)

___
*[`src/test/kotlin/org/wfanet/measurement/common/FlowKtTest.kt` line 115 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/161#-NAQAhGG5HSjQi_AF0D3:-NAQAhGG5HSjQi_AF0D4:boznsgb) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/6500f74bcfe5920ad710fa641c564a28d7c0b7d0/src/test/kotlin/org/wfanet/measurement/common/FlowKtTest.kt#L115)):*
> ```Kotlin
>     runBlocking {
>       val successfullyProcessedItems = mutableListOf<Int>()
>       val triesCounter = mutableMapOf<Int, Int>()
> ```

Consider changing this to 

```
val triesCounter = mutableMapOf<Int, Int>().withDefault{0}
```

This would allow you to simplify the following code:
```
          triesCounter.putIfAbsent(it, 0)
          triesCounter[it] = 1 + (triesCounter[it] ?: 0)
```

renjiezh (commented):
@matthewclegg I added another function retry with error handler, which is more suitable for our Herald usage. With it, we can handle non-retriable error and attempts-exhausted error in the same way. Do you like this idea? If so, do you think we can replace the formal retry function? Seems it is only used by the herald

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/161#-:-NAQH7bS-kPIgia1V1wb:b-v47pdl)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @matthewclegg and @renjiezh)

matthewclegg (changes requested):
> Reviewed 1 of 1 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/161)* status: all files reviewed, 2 unresolved discussions (waiting on @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/common/Flow.kt` line 92 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/161#-NAlUP7w8zl-MgFYUnGS:-NAlUP7w8zl-MgFYUnGT:bvike0z) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/3572d9283930c6b3bdf1f7b51df14e64a8473741/src/main/kotlin/org/wfanet/measurement/common/Flow.kt#L92)):*
> ```Kotlin
>  * Executes [onEachBlock] for each item in the flow. When the block throws a [Throwable] it will
>  * call [errorHandlingBlock] to handle the exception. [errorHandlingBlock] returns Boolean which
>  * indicates whether retry [onEachBlock].
> ```

What happens if [errorHandlingBlock] returns false?  What happens if maxAttempts is exceeded?

See section 6.1 of this article.  https://www.baeldung.com/java-exceptions

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/161)* status: all files reviewed, 2 unresolved discussions (waiting on @matthewclegg)

___
*[`src/main/kotlin/org/wfanet/measurement/common/Flow.kt` line 92 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/161#-NAlUP7w8zl-MgFYUnGS:-NAm0_CrEIIn-VtdRrAa:bftuqxj) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/3572d9283930c6b3bdf1f7b51df14e64a8473741/src/main/kotlin/org/wfanet/measurement/common/Flow.kt#L92)):*
<details><summary><i>Previously, matthewclegg (Matthew Clegg) wrote…</i></summary><blockquote>

What happens if [errorHandlingBlock] returns false?  What happens if maxAttempts is exceeded?

See section 6.1 of this article.  https://www.baeldung.com/java-exceptions

</blockquote></details>

`errorHandlingBlock` returning false will end retries in advance. This is for non-transient errors. (Previously we had `retryPredicate` to determine this.) 
For `errorHandlingBlock(T, Throwable, Bool)`, if an exception catched with`maxAttempts` reached, the bool value passed to `errorHandlingBlock` will be true, then the handling block can take corresponding action (usually regards the error as non-transient).
The modified version will swallow exceptions by running `errorHandlingBlock`. Thus it guarantees no exception re-throw again. Also considering that, the non-transient errors and maxAttempts-reached errors usually have the same solution, so it is easier to use a single  `errorHandlingBlock` to deal with all errors.
___
*[`src/test/kotlin/org/wfanet/measurement/common/FlowKtTest.kt` line 115 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/161#-NAQAhGG5HSjQi_AF0D3:-NAm41lb4fzT8I7f1O3t:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/6500f74bcfe5920ad710fa641c564a28d7c0b7d0/src/test/kotlin/org/wfanet/measurement/common/FlowKtTest.kt#L115)):*
<details><summary><i>Previously, matthewclegg (Matthew Clegg) wrote…</i></summary><blockquote>

Consider changing this to 

```
val triesCounter = mutableMapOf<Int, Int>().withDefault{0}
```

This would allow you to simplify the following code:
```
          triesCounter.putIfAbsent(it, 0)
          triesCounter[it] = 1 + (triesCounter[it] ?: 0)
```



</blockquote></details>

Done.

matthewclegg (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/161)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @renjiezh)

PR #165, opened 2022-09-28
Add options params to buildMutualTlsChannel()

renjiezh:


SanjayVas (changes requested):
> Reviewed 2 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/165)* status: 2 of 3 files reviewed, 2 unresolved discussions (waiting on @renjiezh and @SanjayVas)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/165#-ND5J1GBC8p_Zrb000ZT:-ND5J1GBC8p_Zrb000ZU:b-u9yz6g) (no related file):*
Can you explain why you'd want to do this? I don't know of any channel options we'd want to set globally.

___
___
*[`imports/java/io/netty/transport/BUILD.bazel` line 4 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/165#-ND5IkISEwVVwcDQeQu8:-ND5IkITC1dGKAsbl4DX:b-w3w5cb) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/56677049e924486d78bba74519fa0e70f69168a1/imports/java/io/netty/transport/BUILD.bazel#L4)):*
> ```Starlark
> 
> alias(
>     name = "transport",
> ```

The package in the //imports should match the Java package. That implies that this library only contains symbols in the `io.netty.transport` Java package, but it seems like this is being used to import `io.netty.channel.ChannelOption`.

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/165)* status: 2 of 3 files reviewed, 2 unresolved discussions (waiting on @SanjayVas)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/165#-ND5J1GBC8p_Zrb000ZT:-ND5KmV249dAu722P6v6:btbtm33) (no related file):*

Can you explain why you'd want to do this? I don't know of any channel options we'd want to set globally.
</blockquote></details>

This is for service_config. We need to pass the config via channelOption.
https://www.retinadata.com/blog/configuring-grpc-retries/

___
___
*[`imports/java/io/netty/transport/BUILD.bazel` line 4 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/165#-ND5IkISEwVVwcDQeQu8:-ND5KsHKCk39dU-UUIhQ:b-kgtsc5) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/56677049e924486d78bba74519fa0e70f69168a1/imports/java/io/netty/transport/BUILD.bazel#L4)):*

The package in the //imports should match the Java package. That implies that this library only contains symbols in the `io.netty.transport` Java package, but it seems like this is being used to import `io.netty.channel.ChannelOption`.
</blockquote></details>

`io.netty.channel` is in the package `io.netty.transport`. 
https://github.com/netty/netty/tree/4.1/transport/src/main/java/io/netty/channel
Any alternative way to import it?

SanjayVas (changes requested):
> Reviewed 1 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/165)* status: all files reviewed, 2 unresolved discussions (waiting on @renjiezh)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/165#-ND5J1GBC8p_Zrb000ZT:-ND5aoQB-pgmSGuh17wP:b-1gk6k6) (no related file):*
<details><summary><i>Previously, renjiezh wrote…</i></summary><blockquote>

This is for service_config. We need to pass the config via channelOption.
https://www.retinadata.com/blog/configuring-grpc-retries/
</blockquote></details>

I'm not sure this is the standard way to specify the service config. My understanding is that the config is usually published by the service rather than specified by the client. https://github.com/grpc/grpc/blob/master/doc/service_config.md mentions that the config is usually returned by a name resolver plugin. I'm not sure if that's tricky to set up and/or not applicable to our use case, but that should be investigated.

___
___
*[`imports/java/io/netty/transport/BUILD.bazel` line 4 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/165#-ND5IkISEwVVwcDQeQu8:-ND5YzUABOB8CJuT9a8-:b-rolxh7) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/56677049e924486d78bba74519fa0e70f69168a1/imports/java/io/netty/transport/BUILD.bazel#L4)):*
<details><summary><i>Previously, renjiezh wrote…</i></summary><blockquote>

`io.netty.channel` is in the package `io.netty.transport`. 
https://github.com/netty/netty/tree/4.1/transport/src/main/java/io/netty/channel
Any alternative way to import it?
</blockquote></details>

I think you're mixing up Java package and Maven artifact. `io.netty.channel` is the package.

* If a Maven artifact is the only one which includes symbols for a given Java package, we put the target in the BUILD file under the Bazel package path that matches the Java package and give the target the same name as the Bazel package.
* If there are multiple Maven artifacts that include symbols from the same Java package, then the target for each artifact should be in the BUILD file under the Bazel package path that matches the lowest common Java package. The target name generally references the artifact name in some way, and does *not* take on the name of the Bazel package.

In this case, it looks like the `netty-transport` Maven artifact includes symbols in both `io.netty.channel` and `io.netty.bootstrap`. Therefore, the Bazel package should be //imports/java/io/netty. I'd just call the target `transport`.

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/165)* status: all files reviewed, 1 unresolved discussion (waiting on @renjiezh)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/165#-ND5J1GBC8p_Zrb000ZT:-ND5dAsXCUVDdMdDf1BE:b-aquw5f) (no related file):*

I'm not sure this is the standard way to specify the service config. My understanding is that the config is usually published by the service rather than specified by the client. https://github.com/grpc/grpc/blob/master/doc/service_config.md mentions that the config is usually returned by a name resolver plugin. I'm not sure if that's tricky to set up and/or not applicable to our use case, but that should be investigated.
</blockquote></details>

I did some digging and it sounds like that's all related to service discovery and load balancing. It might be something we should set up in the future, but I suppose for now doing this in the client is the best approach.

___

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/165)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/165#-ND5J1GBC8p_Zrb000ZT:-ND9hkD4AWlDRPe8vpxS:b-c0eoh8) (no related file):*

I did some digging and it sounds like that's all related to service discovery and load balancing. It might be something we should set up in the future, but I suppose for now doing this in the client is the best approach.
</blockquote></details>

Thank you sanjay. The service_config on the client side is like a default value when the server provides none. It looks good to me.

___
___
*[`imports/java/io/netty/transport/BUILD.bazel` line 4 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/165#-ND5IkISEwVVwcDQeQu8:-ND9iV98DsY5KtCdSUMa:b3nfphc) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/56677049e924486d78bba74519fa0e70f69168a1/imports/java/io/netty/transport/BUILD.bazel#L4)):*

I think you're mixing up Java package and Maven artifact. `io.netty.channel` is the package.

* If a Maven artifact is the only one which includes symbols for a given Java package, we put the target in the BUILD file under the Bazel package path that matches the Java package and give the target the same name as the Bazel package.
* If there are multiple Maven artifacts that include symbols from the same Java package, then the target for each artifact should be in the BUILD file under the Bazel package path that matches the lowest common Java package. The target name generally references the artifact name in some way, and does *not* take on the name of the Bazel package.

In this case, it looks like the `netty-transport` Maven artifact includes symbols in both `io.netty.channel` and `io.netty.bootstrap`. Therefore, the Bazel package should be //imports/java/io/netty. I'd just call the target `transport`.
</blockquote></details>

Done.
Thanks for the explanation. I will be careful next time. This PR doesn't need the extra package now.

SanjayVas (approved):
> Reviewed 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/165)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @renjiezh)

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/165)* status: all files reviewed, 1 unresolved discussion (waiting on @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/common/grpc/Channel.kt` line 44 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/165#-NEmjjR27dTNj4sXxS7U:-NEmjjR3AT3lDOir0FU8:b-7v3ipo) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/853d020eb0fbea896b13f20e74a3e5cc1d1fa363/src/main/kotlin/org/wfanet/measurement/common/grpc/Channel.kt#L44)):*
> ```Kotlin
> 
>   if (serviceConfig != null) {
>     channelBuilder.defaultServiceConfig(serviceConfig).enableRetry()
> ```

Apparently retries are enabled by default. This mechanism exists to allow you to disable retries if you want to ignore the service config.

See https://github.com/grpc/proposal/blob/master/A6-client-retries.md#disabling-retries



_Code quote:_
```Kotlin
.enableRetry()
```

PR #167, opened 2022-10-07
remove requirement for GCS flags

jonmolle:
This is used in some cases where the client has a choice of storage and it is confusing to force unused flags to be set.

### DIFF
--- a/src/main/kotlin/org/wfanet/measurement/gcloud/gcs/GcsFromFlags.kt
+++ b/src/main/kotlin/org/wfanet/measurement/gcloud/gcs/GcsFromFlags.kt
@@ -35,7 +35,6 @@ class GcsFromFlags(private val flags: Flags) {
     @CommandLine.Option(
       names = ["--google-cloud-storage-project"],
       description = ["Name of the Google Cloud Storage project to use."],
-      required = true
     )
     lateinit var projectName: String
       private set
@@ -43,7 +42,6 @@ class GcsFromFlags(private val flags: Flags) {
     @CommandLine.Option(
       names = ["--google-cloud-storage-bucket"],
       description = ["Name of the Google Cloud Storage project to use."],
-      required = true
     )
     lateinit var bucket: String
       private set


SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/167)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @jonmolle and @stevenwarejones)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/167#-NDoVxd64x9kp5mGhIZj:-NDoVxd74WDIAqDLYQEJ:b5f157a) (no related file):*
GcsFromFlags.Flags is only intended to be used when GCS is being built from flags. Therefore these flags should always be required.

If there's a use case for conditional usage of GCS, then one should use other flags (e.g. using [argument groups](https://picocli.info/#_argument_groups) to control whether GCS-specific flags are required.

___

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/167)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @jonmolle and @stevenwarejones)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/167#-NDoVxd64x9kp5mGhIZj:-NDoWRTr2RQ32awlHZeU:b4498o3) (no related file):*

GcsFromFlags.Flags is only intended to be used when GCS is being built from flags. Therefore these flags should always be required.

If there's a use case for conditional usage of GCS, then one should use other flags (e.g. using [argument groups](https://picocli.info/#_argument_groups) to control whether GCS-specific flags are required.
</blockquote></details>

A concrete example would be

```kotlin
 @ArgGroup(exclusive = false)
var gcsFlags: GcsFromFlags.Flags? = null
  private set
```

In the above example, gcsFlags will be non-null when either of the flags is set. If either one of the flags is set, the other will be required.

___

jonmolle (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/167)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @jonmolle and @stevenwarejones)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/167#-NDoVxd64x9kp5mGhIZj:-NE2Zy421LCWzVE5W2PB:bedtb0f) (no related file):*

A concrete example would be

```kotlin
 @ArgGroup(exclusive = false)
var gcsFlags: GcsFromFlags.Flags? = null
  private set
```

In the above example, gcsFlags will be non-null when either of the flags is set. If either one of the flags is set, the other will be required.
</blockquote></details>

Good to know, I'll refactor this to something more along those lines. The issue that causes this is where GcsFromFlags is in the code but may never be built, like choosing root storage. Panel Exchange has a storage selector to avoid this issue, perhaps something like that might be more useful generally.

___

PR #179, opened 2023-02-02
Restructure coroutine scopes for Spanner transactions.

SanjayVas:
Breaking API change: TransactionWork no longer takes a TransactionContext parameter, and instead is run with a TransactionScope receiver. This TransactionScope extends CoroutineScope so that CoroutineScope extension functions called from TransactionWork will be run in the TransactionScope.

This also drops the unnecessary transaction timeout, instead expecting callers to implement their own timeouts as necessary.

This is for world-federation-of-advertisers/cross-media-measurement#840

- - -
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/179)

### DIFF
--- a/build/common_jvm_maven.bzl
+++ b/build/common_jvm_maven.bzl
@@ -69,7 +69,7 @@ def common_jvm_maven_artifacts_dict():
         "com.adobe.testing:s3mock-junit4": "2.2.3",
         "com.google.cloud:google-cloud-bigquery": "2.10.10",
         "com.google.cloud:google-cloud-nio": "0.123.28",
-        "com.google.cloud:google-cloud-spanner": "6.34.1",
+        "com.google.cloud:google-cloud-spanner": "6.35.2",
         "com.google.cloud:google-cloud-storage": "2.6.1",
         "com.google.guava:guava": "31.0.1-jre",
         "info.picocli:picocli": "4.4.0",
--- a/src/main/kotlin/org/wfanet/measurement/gcloud/common/Futures.kt
+++ b/src/main/kotlin/org/wfanet/measurement/gcloud/common/Futures.kt
@@ -20,17 +20,11 @@ import com.google.common.util.concurrent.ListenableFuture
 import com.google.common.util.concurrent.MoreExecutors
 import com.google.common.util.concurrent.Uninterruptibles
 import java.util.concurrent.ExecutionException
-import java.util.concurrent.Executor
-import kotlin.coroutines.CoroutineContext
-import kotlin.coroutines.EmptyCoroutineContext
 import kotlin.coroutines.resume
 import kotlin.coroutines.resumeWithException
-import kotlinx.coroutines.CoroutineScope
-import kotlinx.coroutines.CoroutineStart
-import kotlinx.coroutines.guava.future
+import kotlinx.coroutines.Deferred
+import kotlinx.coroutines.guava.asListenableFuture
 import kotlinx.coroutines.suspendCancellableCoroutine
-
-object DirectExecutor : Executor by MoreExecutors.directExecutor()
 
 /**
  * Suspends until the [ApiFuture] completes.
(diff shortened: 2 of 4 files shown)


stevenwarejones (approved):
> Reviewed 3 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/179)* status: 3 of 4 files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/gcloud/common/Futures.kt` line 59 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/179#-NNcLEj85bsFbOnnMF6o:-NNcLEj85bsFbOnnMF6p:burihbt) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/025c6d80135a7eb0781b9963ac539aeee448e978/src/main/kotlin/org/wfanet/measurement/gcloud/common/Futures.kt#L59)):*
> ```Kotlin
> }
> 
> fun <T> Deferred<T>.asApiFuture(): ApiFuture<T> = asListenableFuture().asApiFuture()
> ```

do these belong in common-jvm?

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/179)* status: 3 of 4 files reviewed, 1 unresolved discussion (waiting on @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/gcloud/common/Futures.kt` line 59 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/179#-NNcLEj85bsFbOnnMF6o:-NNcuDKFEsNYvRBCeMz0:bk2k2pn) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/025c6d80135a7eb0781b9963ac539aeee448e978/src/main/kotlin/org/wfanet/measurement/gcloud/common/Futures.kt#L59)):*

do these belong in common-jvm?
</blockquote></details>

I believe these are in the right place (the `gcloud` tree in common-jvm), which is the same place that common Spanner and GCS code lives. ApiFuture is used by multiple gcloud client libraries, though I believe we only interact with them for Spanner (e.g. most our BigQuery code goes through Beam so doesn't use ApiFuture directly).

Did you have somewhere in mind that you think is a better fit?

stevenwarejones (approved):
> Reviewed 1 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/179)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/gcloud/common/Futures.kt` line 59 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/179#-NNcLEj85bsFbOnnMF6o:-NNd-IWiCnu4udNnPl8-:bx7v8kd) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/025c6d80135a7eb0781b9963ac539aeee448e978/src/main/kotlin/org/wfanet/measurement/gcloud/common/Futures.kt#L59)):*

I believe these are in the right place (the `gcloud` tree in common-jvm), which is the same place that common Spanner and GCS code lives. ApiFuture is used by multiple gcloud client libraries, though I believe we only interact with them for Spanner (e.g. most our BigQuery code goes through Beam so doesn't use ApiFuture directly).

Did you have somewhere in mind that you think is a better fit?
</blockquote></details>

If other cloud deploys needed this code then I guess we can move it then.

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/179)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/gcloud/common/Futures.kt` line 59 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/179#-NNcLEj85bsFbOnnMF6o:-NNd-d7dDOg7-xuHtpeP:bny744u) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/025c6d80135a7eb0781b9963ac539aeee448e978/src/main/kotlin/org/wfanet/measurement/gcloud/common/Futures.kt#L59)):*

If other cloud deploys needed this code then I guess we can move it then.
</blockquote></details>

I'm somewhat confused by your comment, as

1. This is in common-jvm in the `gcloud` package, so it should be available for any of our gcloud library code.
2. ApiFuture is type specific to Google Cloud Java client libraries, so it won't be used by code for other cloud providers. It's equivalent to Guava's ListenableFuture, but exists to avoid the gcloud API being publicly coupled to Guava. (We already have a Guava dependency, so we can adapt from ListenableFuture in our code, and then make use of `kotlinx.coroutines.guava`)

PR #182, opened 2023-02-13
Drop TransactionScope, passing TransactionContext as parameter instead.

SanjayVas:
Due to the fact that TransactionScope implemented CoroutineScope, it would be incorrect to use TransactionScope as a receiver for suspending helper functions (see the SuspendFunctionOnCoroutineScope warning). Dropping the type prevents that from occurring.


SanjayVas:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/182)

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/182)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @SanjayVas)

PR #206, opened 2023-07-26
Add functions to convert proto timestamp to date and viceversa

Marco-Premier:
Add function to convert Instant to a Date

[#1123](https://github.com/world-federation-of-advertisers/cross-media-measurement/pull/1123) depends on this

### DIFF
--- a/src/main/kotlin/org/wfanet/measurement/common/ProtoUtils.kt
+++ b/src/main/kotlin/org/wfanet/measurement/common/ProtoUtils.kt
@@ -132,18 +132,13 @@ fun Message.Builder.mergeFromTextProto(textProto: Readable, typeRegistry: TypeRe
 }
 
 /** Converts this [Timestamp] to a [Date]. */
-fun Timestamp.toProtoDate(): Date {
+fun Timestamp.toDateUtc(): Date {
   return toInstant().atZone(ZoneOffset.UTC).toLocalDate().toProtoDate()
 }
 
 /** Converts this [Date] to a [Timestamp]. */
-fun Date.toProtoTime(): Timestamp {
+fun Date.atStartOfDayUtc(): Timestamp {
   return toLocalDate().atStartOfDay().toInstant(ZoneOffset.UTC).toProtoTime()
-}
-
-/** Converts this [Instant] to a [Date] */
-private fun Instant.toDate(): Date {
-  return this.atZone(ZoneOffset.UTC).toLocalDate().toProtoDate()
 }
 
 @Suppress("UNCHECKED_CAST") // Safe per Message contract.
--- a/src/main/kotlin/org/wfanet/measurement/common/ProtoUtils.kt
+++ b/src/main/kotlin/org/wfanet/measurement/common/ProtoUtils.kt
@@ -26,12 +26,14 @@ import com.google.protobuf.TypeRegistry
 import com.google.protobuf.duration
 import com.google.protobuf.timestamp
 import com.google.protobuf.util.JsonFormat
+import com.google.type.Date
 import com.google.type.Interval
 import com.google.type.interval
 import java.io.File
 import java.time.Clock
 import java.time.Duration
 import java.time.Instant
+import java.time.ZoneOffset
 
 /** Converts a protobuf [MessageOrBuilder] into its canonical JSON representation. */
 fun MessageOrBuilder.toJson(): String {
@@ -127,6 +129,21 @@ val ProtocolMessageEnum.numberAsLong: Long
 
 fun Message.Builder.mergeFromTextProto(textProto: Readable, typeRegistry: TypeRegistry) {
   TextFormat.Parser.newBuilder().setTypeRegistry(typeRegistry).build().merge(textProto, this)
+}
+
+/** Converts this [Timestamp] to a [Date]. */
+fun Timestamp.toProtoDate(): Date {
+  return toInstant().atZone(ZoneOffset.UTC).toLocalDate().toProtoDate()
+}
+
+/** Converts this [Date] to a [Timestamp]. */
+fun Date.toProtoTime(): Timestamp {
+  return toLocalDate().atStartOfDay().toInstant(ZoneOffset.UTC).toProtoTime()
+}
+
+/** Converts this [Instant] to a [Date] */
+private fun Instant.toDate(): Date {
+  return this.atZone(ZoneOffset.UTC).toLocalDate().toProtoDate()
 }
 
 @Suppress("UNCHECKED_CAST") // Safe per Message contract.


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/206)

jojijac0b (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/206)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @SanjayVas)

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/206)* status: all files reviewed, 5 unresolved discussions (waiting on @Marco-Premier)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/206#-NaIjbIY3OQ953Xxzytq:-NaIjbIZ9gNGnQCWgQ2T:b-6g0z1b) (no related file):*
I'm not sure I like having these conversion extension methods as it's difficult to convey the assumptions that the methods are making in code. The existing conversion methods force these to be explicit at the call site.

If you really want to avoid the repetition, it's easier for the assumptions to be made more explicit if they convert proto->proto or Java->Java so we don't need to convey that extra information.

* `fun Instant.toLocalDateUtc(): LocalDate`
* `fun LocalDate.atStartOfDayUtc(): Instant`
* `fun Timestamp.toDateUtc(): Date`
* `fun Date.atStartOfDayUtc(): Timestamp`

___
___
*[`src/main/kotlin/org/wfanet/measurement/common/ProtoUtils.kt` line 135 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/206#-NaIjlHW9Aoqg_Ohq5eC:-NaIjlHW9Aoqg_Ohq5eD:b-ni2vny) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/966a26fba01867953749d865b43203e66987b6b1/src/main/kotlin/org/wfanet/measurement/common/ProtoUtils.kt#L135)):*
> ```Kotlin
> 
> /** Converts this [Timestamp] to a [Date]. */
> fun Timestamp.toProtoDate(): Date {
> ```

Make it clear that this is assuming the timezone of the [Timestamp] is in UTC.



_Code quote:_
```Kotlin
toProtoDate
```
___
*[`src/main/kotlin/org/wfanet/measurement/common/ProtoUtils.kt` line 140 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/206#-NaIjrwHBf6JHyZ0CaIo:-NaIjrwHBf6JHyZ0CaIp:b6z524e) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/966a26fba01867953749d865b43203e66987b6b1/src/main/kotlin/org/wfanet/measurement/common/ProtoUtils.kt#L140)):*
> ```Kotlin
> 
> /** Converts this [Date] to a [Timestamp]. */
> fun Date.toProtoTime(): Timestamp {
> ```

Make it clear that this is returning a Timestamp for the start of the day in the UTC timezone.



_Code quote:_
```Kotlin
toProtoTime
```
___
*[`src/main/kotlin/org/wfanet/measurement/common/ProtoUtils.kt` line 145 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/206#-NaIk0Zv4YLH6KeyLc8w:-NaIk0Zv4YLH6KeyLc8x:b-ni2vny) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/966a26fba01867953749d865b43203e66987b6b1/src/main/kotlin/org/wfanet/measurement/common/ProtoUtils.kt#L145)):*
> ```Kotlin
> 
> /** Converts this [Instant] to a [Date] */
> private fun Instant.toDate(): Date {
> ```

Make it clear that this is assuming the timezone of the [Timestamp] is in UTC.



_Code quote:_
```Kotlin
toDate
```
___
*[`src/main/kotlin/org/wfanet/measurement/common/ProtoUtils.kt` line 146 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/206#-NaIlZLU1NSsTqsCbEKD:-NaIlZLU1NSsTqsCbEKE:b-ww8una) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/966a26fba01867953749d865b43203e66987b6b1/src/main/kotlin/org/wfanet/measurement/common/ProtoUtils.kt#L146)):*
> ```Kotlin
> /** Converts this [Instant] to a [Date] */
> private fun Instant.toDate(): Date {
>   return this.atZone(ZoneOffset.UTC).toLocalDate().toProtoDate()
> ```

Redundant `this`



_Code quote:_
```Kotlin
this.
```

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/206)* status: all files reviewed, 5 unresolved discussions (waiting on @Marco-Premier)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/206#-NaIjbIY3OQ953Xxzytq:-NaIoFS1AzgEO_HmM2os:bovdwx1) (no related file):*

I'm not sure I like having these conversion extension methods as it's difficult to convey the assumptions that the methods are making in code. The existing conversion methods force these to be explicit at the call site.

If you really want to avoid the repetition, it's easier for the assumptions to be made more explicit if they convert proto->proto or Java->Java so we don't need to convey that extra information.

* `fun Instant.toLocalDateUtc(): LocalDate`
* `fun LocalDate.atStartOfDayUtc(): Instant`
* `fun Timestamp.toDateUtc(): Date`
* `fun Date.atStartOfDayUtc(): Timestamp`
</blockquote></details>

Note that even with the above methods, there's still some inconsistency in the naming. The `Utc` suffix for the `at` methods is correctly implying that it's picking a time within the day. The `Utc` suffix for the `to` methods is weirder, because there it's actually indicating an assumption about the input time but appears that it's applying the `Utc` modifier to the output date.

___

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/206)* status: 0 of 1 files reviewed, 5 unresolved discussions (waiting on @jojijac0b, @Marco-Premier, and @SanjayVas)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/206#-NaIjbIY3OQ953Xxzytq:-NaLBe_d5ANSvLlXVcTZ:bprv5qx) (no related file):*

Note that even with the above methods, there's still some inconsistency in the naming. The `Utc` suffix for the `at` methods is correctly implying that it's picking a time within the day. The `Utc` suffix for the `to` methods is weirder, because there it's actually indicating an assumption about the input time but appears that it's applying the `Utc` modifier to the output date.
</blockquote></details>

I see your point @SanjayVas . I removed the Instant to Date function entirely as it's supposed to be used only for testing and I can avoid the conversion entirely by declaring variable of type ```Date``` instead of ```Instant```here (https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/src/test/kotlin/org/wfanet/measurement/api/v2alpha/tools/MeasurementSystemTest.kt#L244) , which also makes  thing more readable.

Regarding the ````fun Timestamp.toDateUtc(): Date``` naming inconsistency, I'm not sure what name could clarify that ambiguity, maybe: ```Timestamp.asUtcToDate``` ? This would make it clearer that it's the Timestamp being modify as UTC and then eventually to Date.

What you think?

___

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/206)* status: 0 of 1 files reviewed, 5 unresolved discussions (waiting on @jojijac0b and @Marco-Premier)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/206#-NaIjbIY3OQ953Xxzytq:-NaNHv6R91jYOzU4CfMV:b-a92twq) (no related file):*
> maybe: `Timestamp.asUtcToDate`? 

The `as` prefix has its own meaning. It's used by convention for methods that return a view (such that if the original object changed, the view would reflect that), as opposed to the `to` prefix for methods that give you an object that does not remain tied to the original. Putting both `as` and `to` in a method name is quite confusing.

> This would make it clearer that it's the Timestamp being modify as UTC and then eventually to Date.

My point was that the existing conversion methods make this very clear already. The only benefit of combining any of them together is to avoid repetition if we tend to frequently chain the same operations together. I don't think it's worth reducing that level of repetition at the cost of clarity.

___

PR #209, opened 2023-08-03
Jojijacob signed store

jojijac0b:
Creates signed store to be used to upload and download signed blobs

### DIFF
--- a/src/main/kotlin/org/wfanet/measurement/common/crypto/SignedStore.kt
+++ b/src/main/kotlin/org/wfanet/measurement/common/crypto/SignedStore.kt
@@ -1,0 +1,76 @@
+// Copyright 2021 The Cross-Media Measurement Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package org.wfanet.measurement.common.crypto
+
+import com.google.protobuf.ByteString
+import java.security.PrivateKey
+import com.google.protobuf.kotlin.toByteString
+import java.security.cert.X509Certificate
+import org.wfanet.measurement.common.flatten
+import org.wfanet.measurement.storage.StorageClient
+
+/** Store of blob and signature. */
+class SignedStore(storageClient: StorageClient) {
+  private val storageClient = storageClient
+
+  internal fun blobKeyForSignature(blobKey: String): String {
+    return "signature.$blobKey"
+  }
+
+  internal fun blobKeyForContent(blobKey: String): String {
+    return "content.$blobKey"
+  }
+
+  internal suspend fun write(blobKey: String, x509: X509Certificate, privateKey: PrivateKey, content: Flow<ByteString>): String {
+    // Since StorageClient has no concept of "overwriting" a blob, we first delete existing blobs.
+    // This is to ensure that transient failures after some blobs are written do not cause problems
+    // when re-attempting to write.
+    storageClient.getBlob(blobKeyForContent(blobKey))?.delete()
+    storageClient.getBlob(blobKeyForSignature(blobKey))?.delete()
+
+    val signedBlob = content.onEach(privateKey.newSigner(x509)::update)
+    val signature = privateKey.newSigner(x509).sign().toByteString()
+    storageClient.writeBlob(blobKeyForContent(blobKey), signedBlob)
+    storageClient.writeBlob(blobKeyForSignature(blobKey), signature)
+    return blobKey
+  }
+
+  @Throws(BlobNotFoundException::class)
+  suspend fun read(blobKey: String, x509: X509Certificate): VerifiedBlob? {
+    val signedBlob = storageClient.getBlob(blobKeyForContent(blobKey)) ?: throw BlobNotFoundException(blobKey)
+    val signature = storageClient.getBlob(blobKeyForSignature(blobKey))
+    return VerifiedBlob(SignedBlob(signedBlob, signature), x509)
+  }
+
+
+  class BlobNotFoundException(inputKey: String) : Exception("$inputKey not found")
+
+  class VerifiedBlob(private val sourceBlob: SignedBlob, private val x509: X509Certificate) {
+    val size: Long
+      get() = sourceBlob.size
+
+    val signature: ByteString
+      get() = sourceBlob.signature
+
+    /** Reads the underlying blob. Throws if the signature was invalid. */
+    fun read(): Flow<ByteString> = sourceBlob.readVerifying(x509)
+
+    /** @see [StorageClient::toByteString]. */
+    suspend fun toByteString(): ByteString = read().flatten()
+
+    /** Reads the blob into a UTF8 String. */
+    suspend fun toStringUtf8(): String = toByteString().toStringUtf8()
+  }
+}
--- a/src/main/kotlin/org/wfanet/measurement/common/crypto/tink/DownloadClient.kt
+++ b/src/main/kotlin/org/wfanet/measurement/common/crypto/tink/DownloadClient.kt
@@ -1,0 +1,45 @@
+// Copyright 2022 The Cross-Media Measurement Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package org.wfanet.panelmatch.client.storage
+
+
+import java.security.cert.X509Certificate
+import kotlinx.coroutines.flow.Flow
+import org.wfanet.measurement.common.crypto.SignedBlob
+import org.wfanet.measurement.storage.StorageClient
+import org.wfanet.panelmatch.client.storage.VerifyingStorageClient.VerifiedBlob
+import org.wfanet.panelmatch.protocol.NamedSignature
+
+/**
+ * Interface for that downloads blob from shared storage.
+ */
+interface DownloadClient {
+
+  /**
+   * Reads [blobKey] from shared storage. Returns a [VerifiedBlob] which checks the blob's signature
+   * using [x509] when it is read.
+   *
+   * Note that the final signature validation does not happen until the [Flow] underlying the
+   * [VerifiedBlob] is collected.
+   */
+  @Throws(BlobNotFoundException::class)
+  suspend fun getBlob(sharedStorage: StorageClient, blobKey: String, x509: X509Certificate): VerifiedBlob
+
+  suspend fun verifyBlob(
+    signedBlob: SignedBlob,
+    namedSignature: NamedSignature,
+    x509: X509Certificate
+  ): VerifiedBlob
+}
--- a/src/main/kotlin/org/wfanet/measurement/common/crypto/tink/HybridDecryptor.kt
+++ b/src/main/kotlin/org/wfanet/measurement/common/crypto/tink/HybridDecryptor.kt
@@ -1,0 +1,30 @@
+package org.wfanet.panelmatch.client.storage
+
+import com.google.crypto.tink.HybridDecrypt
+import com.google.crypto.tink.KeysetHandle
+import com.google.protobuf.ByteString
+import kotlinx.coroutines.flow.Flow
+import org.wfanet.measurement.common.crypto.KeyBlobStore
+import org.wfanet.measurement.common.crypto.PrivateKeyStore.KeyId
+
+
+interface HybridDecryptor {
+  /**
+   * Takes in an input and keyID and returns a decrypted Bytestring using HybridDecrypt.
+   *
+   * The keyId is used to retrieve the public key from a PrivateKeyStore. This is used to generate the KeysetHandle
+   */
+  suspend fun hybridDecrypt(
+    input: ByteString,
+    keyBlobStore: KeyBlobStore,
+    keyId: KeyId,
+  ): ByteString
+
+  /**
+   * Given a keyId, returns the appropriate KeysetHandle
+   */
+  suspend fun getPrivateKeysetHandle(
+    keyBlobStore: KeyBlobStore,
+    keyId: KeyId
+  ): KeysetHandle
+}
--- a/src/main/kotlin/org/wfanet/measurement/common/crypto/tink/HybridEncryptor.kt
+++ b/src/main/kotlin/org/wfanet/measurement/common/crypto/tink/HybridEncryptor.kt
@@ -1,0 +1,28 @@
+package org.wfanet.panelmatch.client.storage
+
+import com.google.crypto.tink.KeysetHandle
+import com.google.protobuf.ByteString
+import org.wfanet.measurement.common.crypto.KeyBlobStore
+import org.wfanet.measurement.common.crypto.PrivateKeyStore.KeyId
+
+
+interface HybridEncryptor {
+  /**
+   * Takes in an input and keyID and returns an encrypted Bytestring using HybridEncrypt.
+   *
+   * The keyId is used to retrieve the public key from a PrivateKeyStore. This is used to generate the KeysetHandle
+   */
+  suspend fun hybridEncrypt(
+    input: ByteString,
+    keyBlobStore: KeyBlobStore,
+    keyId: KeyId,
+  ): ByteString
+
+  /**
+   * Given a keyId, returns the appropriate KeysetHandle
+   */
+  suspend fun getPublicKeysetHandle(
+    keyBlobStore: KeyBlobStore,
+    keyId: KeyId
+  ): KeysetHandle
+}
--- a/src/main/kotlin/org/wfanet/measurement/common/crypto/tink/UploadStorageClient.kt
+++ b/src/main/kotlin/org/wfanet/measurement/common/crypto/tink/UploadStorageClient.kt
@@ -1,0 +1,50 @@
+// Copyright 2022 The Cross-Media Measurement Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package org.wfanet.panelmatch.client.storage
+
+import com.google.protobuf.ByteString
+import java.security.PrivateKey
+import java.security.cert.X509Certificate
+import kotlinx.coroutines.flow.Flow
+import org.wfanet.measurement.storage.StorageClient
+import org.wfanet.panelmatch.protocol.NamedSignature
+import java.io.Serializable
+import org.wfanet.measurement.common.crypto.SignedBlob
+
+
+/**
+ * Interface for [StorageClient] that writes (blob, signature) pairs to shared storage.
+ */
+interface UploadStorageClient {
+
+  /**
+   * Writes [content] to shared storage using the given [blobKey]. Also writes a serialized
+   * [NamedSignature] for the blob.
+   */
+  suspend fun writeBlob(
+    sharedStorage: StorageClient,
+    blobKey: String,
+    content: Flow<ByteString>,
+    x509: X509Certificate,
+    privateKey: PrivateKey
+  ): Serializable
+
+  suspend fun signBlob(
+    content: Flow<ByteString>,
+    x509: X509Certificate,
+    privateKey: PrivateKey
+  ): SignedBlob
+
+}


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209)

stevenwarejones (changes requested):
> Reviewed 1 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209)* status: 1 of 5 files reviewed, 5 unresolved discussions (waiting on @jojijac0b)

___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/SignedStore.kt` line 25 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209#-NavI5ptEG_SRGdeY6tn:-NavI5ptEG_SRGdeY6to:b-lcyfz5) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/e7257d3b56fbaac2e376cd1baec5847f152c90ad/src/main/kotlin/org/wfanet/measurement/common/crypto/SignedStore.kt#L25)):*
> ```Kotlin
> 
> /** Store of blob and signature. */
> class SignedStore(storageClient: StorageClient) {
> ```

you should just be able to do 
```
class SignedStore(private val storageClient: StorageClient) {
```
and delete the following line
___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/SignedStore.kt` line 44 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209#-NavK6vKDGiIJtCwRGr_:-NavK6vKDGiIJtCwRGra:b-idrfhn) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/e7257d3b56fbaac2e376cd1baec5847f152c90ad/src/main/kotlin/org/wfanet/measurement/common/crypto/SignedStore.kt#L44)):*
> ```Kotlin
> 
>     val signedBlob = content.onEach(privateKey.newSigner(x509)::update)
>     val signature = privateKey.newSigner(x509).sign().toByteString()
> ```

generate the signature like here:
https://github.com/world-federation-of-advertisers/common-jvm/blob/8f5eefcb821535e870266bd972b55594282d51a4/src/test/kotlin/org/wfanet/measurement/common/crypto/SignaturesTest.kt#L93
___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/SignedStore.kt` line 45 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209#-NavJkvK9N5sT14Rwzha:-NavJkvK9N5sT14Rwzhb:bmz7cu5) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/e7257d3b56fbaac2e376cd1baec5847f152c90ad/src/main/kotlin/org/wfanet/measurement/common/crypto/SignedStore.kt#L45)):*
> ```Kotlin
>     val signedBlob = content.onEach(privateKey.newSigner(x509)::update)
>     val signature = privateKey.newSigner(x509).sign().toByteString()
>     storageClient.writeBlob(blobKeyForContent(blobKey), signedBlob)
> ```

there's not a concept of the `signedBlob` here --- you can just write the content out
___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/SignedStore.kt` line 51 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209#-NavN-_C2GEVn91fPuTE:-NavN-_C2GEVn91fPuTF:baou2ni) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/e7257d3b56fbaac2e376cd1baec5847f152c90ad/src/main/kotlin/org/wfanet/measurement/common/crypto/SignedStore.kt#L51)):*
> ```Kotlin
> 
>   @Throws(BlobNotFoundException::class)
>   suspend fun read(blobKey: String, x509: X509Certificate): VerifiedBlob? {
> ```

i don't think you need `VerifiedBlob` - just have this return a `SignedBlob`

https://github.com/world-federation-of-advertisers/common-jvm/blob/8f5eefcb821535e870266bd972b55594282d51a4/src/main/kotlin/org/wfanet/measurement/common/crypto/SignedBlob.kt#L25
___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/SignedStore.kt` line 52 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209#-NavKFv65yphE9BTc4Vg:-NavKFv65yphE9BTc4Vh:b-syvlxg) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/e7257d3b56fbaac2e376cd1baec5847f152c90ad/src/main/kotlin/org/wfanet/measurement/common/crypto/SignedStore.kt#L52)):*
> ```Kotlin
>   @Throws(BlobNotFoundException::class)
>   suspend fun read(blobKey: String, x509: X509Certificate): VerifiedBlob? {
>     val signedBlob = storageClient.getBlob(blobKeyForContent(blobKey)) ?: throw BlobNotFoundException(blobKey)
> ```

`val content = `

jojijac0b (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209)* status: 1 of 5 files reviewed, 5 unresolved discussions (waiting on @Marco-Premier and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/SignedStore.kt` line 44 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209#-NavK6vKDGiIJtCwRGr_:-NawOFIi41TAiNTI2Ge5:b-ake38r) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/e7257d3b56fbaac2e376cd1baec5847f152c90ad/src/main/kotlin/org/wfanet/measurement/common/crypto/SignedStore.kt#L44)):*

generate the signature like here:
https://github.com/world-federation-of-advertisers/common-jvm/blob/8f5eefcb821535e870266bd972b55594282d51a4/src/test/kotlin/org/wfanet/measurement/common/crypto/SignaturesTest.kt#L93
</blockquote></details>

whats the reasoning to use the collectAndSign method instead of just sign like its done here: 

    val signature = privateKey.sign(certificate, DATA)

jojijac0b (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209)* status: 1 of 5 files reviewed, 5 unresolved discussions (waiting on @Marco-Premier and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/SignedStore.kt` line 51 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209#-NavN-_C2GEVn91fPuTE:-NawPD28Bm9tsr0526dr:bqblk4a) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/e7257d3b56fbaac2e376cd1baec5847f152c90ad/src/main/kotlin/org/wfanet/measurement/common/crypto/SignedStore.kt#L51)):*

i don't think you need `VerifiedBlob` - just have this return a `SignedBlob`

https://github.com/world-federation-of-advertisers/common-jvm/blob/8f5eefcb821535e870266bd972b55594282d51a4/src/main/kotlin/org/wfanet/measurement/common/crypto/SignedBlob.kt#L25
</blockquote></details>

would this verify that the content has not been corrupted?

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209)* status: 1 of 5 files reviewed, 5 unresolved discussions (waiting on @jojijac0b and @Marco-Premier)

___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/SignedStore.kt` line 44 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209#-NavK6vKDGiIJtCwRGr_:-NawpD5Y5N0NCCTXiWT2:brlsfh) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/e7257d3b56fbaac2e376cd1baec5847f152c90ad/src/main/kotlin/org/wfanet/measurement/common/crypto/SignedStore.kt#L44)):*
<details><summary><i>Previously, jojijac0b wrote…</i></summary><blockquote>

whats the reasoning to use the collectAndSign method instead of just sign like its done here: 

    val signature = privateKey.sign(certificate, DATA)
</blockquote></details>

you still need to pass in the data to sign
https://github.com/world-federation-of-advertisers/common-jvm/blob/d6e938ee3c94030f0759695666c11be7a8768128/src/main/kotlin/org/wfanet/measurement/common/crypto/Signatures.kt#L55
___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/SignedStore.kt` line 51 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209#-NavN-_C2GEVn91fPuTE:-NawoJpe-C4Po_DDakTm:bcsfw7d) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/e7257d3b56fbaac2e376cd1baec5847f152c90ad/src/main/kotlin/org/wfanet/measurement/common/crypto/SignedStore.kt#L51)):*
<details><summary><i>Previously, jojijac0b wrote…</i></summary><blockquote>

would this verify that the content has not been corrupted?
</blockquote></details>

this verifies it as it reads it -- https://github.com/world-federation-of-advertisers/common-jvm/blob/d6e938ee3c94030f0759695666c11be7a8768128/src/main/kotlin/org/wfanet/measurement/common/crypto/SignedBlob.kt#L45C7-L45C20

jojijac0b (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209)* status: 0 of 6 files reviewed, 5 unresolved discussions (waiting on @Marco-Premier and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/SignedStore.kt` line 25 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209#-NavI5ptEG_SRGdeY6tn:-NbFQUDL67A1_L85Q_Dk:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/e7257d3b56fbaac2e376cd1baec5847f152c90ad/src/main/kotlin/org/wfanet/measurement/common/crypto/SignedStore.kt#L25)):*

you should just be able to do 
```
class SignedStore(private val storageClient: StorageClient) {
```
and delete the following line
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/SignedStore.kt` line 44 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209#-NavK6vKDGiIJtCwRGr_:-NbFQYDjC8WCT2yEp168:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/e7257d3b56fbaac2e376cd1baec5847f152c90ad/src/main/kotlin/org/wfanet/measurement/common/crypto/SignedStore.kt#L44)):*

you still need to pass in the data to sign
https://github.com/world-federation-of-advertisers/common-jvm/blob/d6e938ee3c94030f0759695666c11be7a8768128/src/main/kotlin/org/wfanet/measurement/common/crypto/Signatures.kt#L55
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/SignedStore.kt` line 45 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209#-NavJkvK9N5sT14Rwzha:-NbFQXPODWcl3ljZDsvX:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/e7257d3b56fbaac2e376cd1baec5847f152c90ad/src/main/kotlin/org/wfanet/measurement/common/crypto/SignedStore.kt#L45)):*

there's not a concept of the `signedBlob` here --- you can just write the content out
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/SignedStore.kt` line 51 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209#-NavN-_C2GEVn91fPuTE:-NbFQVjqAHnh3zDzXZSG:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/e7257d3b56fbaac2e376cd1baec5847f152c90ad/src/main/kotlin/org/wfanet/measurement/common/crypto/SignedStore.kt#L51)):*

this verifies it as it reads it -- https://github.com/world-federation-of-advertisers/common-jvm/blob/d6e938ee3c94030f0759695666c11be7a8768128/src/main/kotlin/org/wfanet/measurement/common/crypto/SignedBlob.kt#L45C7-L45C20
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/SignedStore.kt` line 52 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209#-NavKFv65yphE9BTc4Vg:-NbFQVMP7SKy2UDjSE3a:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/e7257d3b56fbaac2e376cd1baec5847f152c90ad/src/main/kotlin/org/wfanet/measurement/common/crypto/SignedStore.kt#L52)):*

`val content = `
</blockquote></details>

Done.

jojijac0b (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209)* status: 0 of 6 files reviewed, 14 unresolved discussions (waiting on @Marco-Premier and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/EncryptedSignedStore.kt` line 16 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209#-NbGZ5Gu5xgHyRAlj4Ng:-NbH5R4m8Pq1JmrTvF01:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/75b7ad4132826701f9c2d3abddeb8b09906845a1/src/main/kotlin/org/wfanet/measurement/common/crypto/EncryptedSignedStore.kt#L16)):*

i don't think write should be internal
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/EncryptedSignedStore.kt` line 23 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209#-NbGc--40zbqNQHzlPcw:-NbH5Rcv05ChZKR-Z5un:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/75b7ad4132826701f9c2d3abddeb8b09906845a1/src/main/kotlin/org/wfanet/measurement/common/crypto/EncryptedSignedStore.kt#L23)):*

you can just do 
```
val encryptedData = publicKeysetHandle.hybridEncrypt(content, null)
```

</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/EncryptedSignedStore.kt` line 27 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209#-NbGZ4U71E_g46mw2J0q:-NbH5SjfACsfJtybLHRF:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/75b7ad4132826701f9c2d3abddeb8b09906845a1/src/main/kotlin/org/wfanet/measurement/common/crypto/EncryptedSignedStore.kt#L27)):*

ditto
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/EncryptedSignedStore.kt` line 32 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209#-NbGcGWW554uuvyEP7dL:-NbH5e580FTTKp-89GLR:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/75b7ad4132826701f9c2d3abddeb8b09906845a1/src/main/kotlin/org/wfanet/measurement/common/crypto/EncryptedSignedStore.kt#L32)):*

this shouldn't compile in its current state.... see my other content about changing the return type of `SignedStore.read`

Also, you'll need to collect the flow here too. and then return it `asFlow`
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/EncryptedSignedStore.kt` line 33 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209#-NbGc2iZ0N3ouTZ8Q-gA:-NbH5VQm7rJUj0ppSfTw:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/75b7ad4132826701f9c2d3abddeb8b09906845a1/src/main/kotlin/org/wfanet/measurement/common/crypto/EncryptedSignedStore.kt#L33)):*

ditto on no need to pass null
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/SignedStore.kt` line 25 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209#-NbGZG3E6OfwHfMSD2Kn:-NbH5fGq4Aw418ZUmWnE:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/75b7ad4132826701f9c2d3abddeb8b09906845a1/src/main/kotlin/org/wfanet/measurement/common/crypto/SignedStore.kt#L25)):*

I'd make this private instead of internal. I'd also prefer 
```
return "signature/$blobKey"
```

</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/SignedStore.kt` line 29 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209#-NbGZIhR9b7rwu9qwO1e:-NbH5foX00GWNus9g6qw:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/75b7ad4132826701f9c2d3abddeb8b09906845a1/src/main/kotlin/org/wfanet/measurement/common/crypto/SignedStore.kt#L29)):*

ditto on making it private but use `/` instead of `.`
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/SignedStore.kt` line 33 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209#-NbGZfMR7sfgNkGP39R3:-NbH5git-Hqk_HmEw6QN:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/75b7ad4132826701f9c2d3abddeb8b09906845a1/src/main/kotlin/org/wfanet/measurement/common/crypto/SignedStore.kt#L33)):*

i think this, too, shouldn't be internal
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/SignedStore.kt` line 52 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209#-NbGZiICCIPLR0p0sqJT:-NbH5hV3-WgJq7-9XQRj:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/75b7ad4132826701f9c2d3abddeb8b09906845a1/src/main/kotlin/org/wfanet/measurement/common/crypto/SignedStore.kt#L52)):*

i think this should be public. Also - wdyt about just having this return a `Flow<ByteString` directly?
```
return SignedBlob(content, signature).readVerifying(x509)
```

</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/SignedStore.kt` line 52 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209#-NbGbSti6PW_5NW6g2tB:-NbH5jr_AQbPAOY7VHbO:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/75b7ad4132826701f9c2d3abddeb8b09906845a1/src/main/kotlin/org/wfanet/measurement/common/crypto/SignedStore.kt#L52)):*

return type should just be `SignedBlob`
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/tink/DownloadClient.kt` line 28 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209#-NbGdS2F3Uu-ZEfcLxCp:-NbH5nSs8ZpHrzjDL1u3:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/75b7ad4132826701f9c2d3abddeb8b09906845a1/src/main/kotlin/org/wfanet/measurement/common/crypto/tink/DownloadClient.kt#L28)):*

delete this file
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/tink/HybridDecryptor.kt` line 11 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209#-NbGdTNH0uTizoCfLzyZ:-NbH5oKxB15EmlEURXtq:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/75b7ad4132826701f9c2d3abddeb8b09906845a1/src/main/kotlin/org/wfanet/measurement/common/crypto/tink/HybridDecryptor.kt#L11)):*

delete this file
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/tink/HybridEncryptor.kt` line 15 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209#-NbGbnFL0e3Z-zUPHqNi:-NbH5oxL1WZCN-KdektO:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/75b7ad4132826701f9c2d3abddeb8b09906845a1/src/main/kotlin/org/wfanet/measurement/common/crypto/tink/HybridEncryptor.kt#L15)):*

delete this file
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/tink/UploadStorageClient.kt` line 30 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209#-NbGbkhH6CFJ5B8ypjtU:-NbH5poeCKRnN5X3Wmkx:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/75b7ad4132826701f9c2d3abddeb8b09906845a1/src/main/kotlin/org/wfanet/measurement/common/crypto/tink/UploadStorageClient.kt#L30)):*

delete this file
</blockquote></details>

Done.

stevenwarejones (changes requested):
Needs tests too.

> Reviewed 6 of 6 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209#-:-NbKerzi2WtGATI0uBdZ:b-px5ev3)* status: all files reviewed, 3 unresolved discussions (waiting on @jojijac0b and @Marco-Premier)

___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/SignedStore.kt` line 25 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209#-NbGZG3E6OfwHfMSD2Kn:-NbKeQwV6owZu5X4veel:b-oq69aj) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/75b7ad4132826701f9c2d3abddeb8b09906845a1/src/main/kotlin/org/wfanet/measurement/common/crypto/SignedStore.kt#L25)):*
<details><summary><i>Previously, jojijac0b wrote…</i></summary><blockquote>

Done.
</blockquote></details>

can you make it private?
___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/SignedStore.kt` line 29 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209#-NbGZIhR9b7rwu9qwO1e:-NbKeUTV6NbS5VOsAcbA:b-oq69aj) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/75b7ad4132826701f9c2d3abddeb8b09906845a1/src/main/kotlin/org/wfanet/measurement/common/crypto/SignedStore.kt#L29)):*
<details><summary><i>Previously, jojijac0b wrote…</i></summary><blockquote>

Done.
</blockquote></details>

can you make it private?
___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/SignedStore.kt` line 52 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209#-NbKeYUJCJk_S-KLuIVa:-NbKeYUJCJk_S-KLuIVb:buqcuvr) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/357c235aaf5391a8514d82c8eae3bcbf3c071a29/src/main/kotlin/org/wfanet/measurement/common/crypto/SignedStore.kt#L52)):*
> ```Kotlin
> 
>   @Throws(BlobNotFoundException::class)
>   suspend fun read(blobKey: String, x509: X509Certificate): Flow<ByteString>? {
> ```

you don't need to return a nullable flow I don't think.... just `Flow<ByteString>`

stevenwarejones (changes requested):
Needs tests still.

> Reviewed 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209#-:-NbzhDcD8AYYL9zRBDHS:blioctp)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Marco-Premier)

stevenwarejones (changes requested):
> Reviewed 6 of 6 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209)* status: all files reviewed, 2 unresolved discussions (waiting on @jojijac0b, @Marco-Premier, and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/EncryptedSignedStore.kt` line 53 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209#-Nc4SLDBAapP8ZIVzUkL:-Nc4SLDBAapP8ZIVzUkM:b4nk90l) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/cb499f5c87db109127eaf265c7ad4f169bcaa700/src/main/kotlin/org/wfanet/measurement/common/crypto/EncryptedSignedStore.kt#L53)):*
> ```Kotlin
>   }
> 
>   suspend fun read(
> ```

this should throw blobnotfound exception
___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/EncryptedSignedStore.kt` line 57 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209#-Nc4SER5DznrmBUYrjMu:-Nc4SER5DznrmBUYrjMv:b9v0dbh) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/cb499f5c87db109127eaf265c7ad4f169bcaa700/src/main/kotlin/org/wfanet/measurement/common/crypto/EncryptedSignedStore.kt#L57)):*
> ```Kotlin
>     signingX509: X509Certificate,
>     decryptingPrivateKeyHandle: PrivateKeyHandle
>   ): Flow<ByteString>? {
> ```

you should need the nullable param here -- just `Flow<ByteString>`

jojijac0b (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209)* status: 9 of 10 files reviewed, 2 unresolved discussions (waiting on @Marco-Premier, @SanjayVas, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/EncryptedSignedStore.kt` line 53 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209#-Nc4SLDBAapP8ZIVzUkL:-Nc8D9tD5_OY8obeEEtY:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/cb499f5c87db109127eaf265c7ad4f169bcaa700/src/main/kotlin/org/wfanet/measurement/common/crypto/EncryptedSignedStore.kt#L53)):*

this should throw blobnotfound exception
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/EncryptedSignedStore.kt` line 57 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209#-Nc4SER5DznrmBUYrjMu:-Nc8DAvcD6uVPSH7iEce:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/cb499f5c87db109127eaf265c7ad4f169bcaa700/src/main/kotlin/org/wfanet/measurement/common/crypto/EncryptedSignedStore.kt#L57)):*

you should need the nullable param here -- just `Flow<ByteString>`
</blockquote></details>

Done.

jojijac0b (commented):
Tests are there. I made an update right after the initial push that adds the test to fix the lint errors so it may not have shown up. Tests are SignedStoreTest and EncryptedSignedStoreTest

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209#-:-Nc8DD9x48480Xi77uU4:b63nwvx)* status: 9 of 10 files reviewed, 2 unresolved discussions (waiting on @Marco-Premier, @SanjayVas, and @stevenwarejones)

stevenwarejones (approved):
> Reviewed 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Marco-Premier and @SanjayVas)

SanjayVas (changes requested):
> Reviewed 3 of 6 files at r5, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209)* status: all files reviewed, 5 unresolved discussions (waiting on @jojijac0b and @Marco-Premier)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209#-Nc9-2Yr2bGNC3HmcLD0:-Nc9-2Yr2bGNC3HmcLD1:b5aad1g) (no related file):*
I'm still trying to understand the use cases to determine whether this is using the current storage abstractions correctly, or if we instead need to change the abstractions to fit the new use case.

A reminder of the current types:

* `StorageClient` - Interface for raw storage access. The intent was not to use this directly but rather access it through a store.
   * `StorageClient.Blob` - Interface that provides a reference to a written blob for a given `StorageClient` implementation.
* `Store` - Abstract class which wraps a `StorageClient` for a specific blob type. It provides two features: private blob key prefix and deterministic blob key derivation.
   * Private blob key prefixes are intended to ensure that the only way a blob is accessed is through the `Store` implementation. This assumes that blobs are always written and read by the same entity, as that was the only use case when the interface was designed.
   * Deterministic blob key derivation is to ensure that you always get the same blob key for the same context (e.g. a Requisition). If you actually just want to use some unique identifier directly then you can just specify the context type as `String`.
* `KmsStorageClient` - A `StorageClient` which wraps another `StorageClient`, passively ensuring blobs are encrypted at rest with a single symmetric key managed by a KMS.
* `SignedBlob` - `StorageClient.Blob` implementation which wraps another `StorageClient.Blob` and a signature, providing methods for verifying that the blob matches the signature.
* `StorageClient.createSignedBlob` - Extension method for creating a `SignedBlob`, i.e. generating a signature while writing a blob.

If we want to handle cases where blobs are intended to be accessible by entities that may not be using Halo code (e.g. a `DataProvider` or `ModelProvider`), then we may want to rethink the `Store` interface a bit or at least rename it to something like `PrefixStore`. Ideally we could still have some `Store` interface so we could have some consistency around anything that abstracts a `StorageClient`, but I understand if that might too far out of scope for this change.

___
___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/EncryptedSignedStore.kt` line 32 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209#-Nc8tR5s2X50_ALNpxnz:-Nc8t6IZ2y0qKz8ZH3Zr:bilutoq) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/766ed7840600cb1ae43ead377fd8ef26171ba198/src/main/kotlin/org/wfanet/measurement/common/crypto/EncryptedSignedStore.kt#L32)):*
> ```Kotlin
> private const val WRITE_BUFFER_SIZE = BYTES_PER_MIB * 5
> 
> class EncryptedSignedStore(private val storageClient: StorageClient) {
> ```

Add KDoc documentation here.

I assume this is a different use case than `KmsStorageClient` which uses symmetric encryption and the same key for all blobs. This use of asymmetric encryption implies that the entity writing the blob is never the entity reading the blob.
___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/SignedStore.kt` line 24 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209#-Nc8zpmb8Bmx7ZHmfDcw:-Nc8zpmc7agY6-cqpV-H:b-13i4gt) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/766ed7840600cb1ae43ead377fd8ef26171ba198/src/main/kotlin/org/wfanet/measurement/common/crypto/SignedStore.kt#L24)):*
> ```Kotlin
> import org.wfanet.measurement.storage.StorageClient
> 
> /** Store of blob and signature. */
> ```

nit: Keep the abstractions at the right level. At the layer at which we're talking about both the content and the signature they are two separate blobs. At the layer that isn't aware of the signature we can treat it as one logical blob.



_Code quote:_
```Kotlin
Store of blob and signature.
```
___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/SignedStore.kt` line 41 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209#-Nc8sBzF90MbXKHj38RV:-Nc8sBzF90MbXKHj38RW:b-qeeqfp) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/766ed7840600cb1ae43ead377fd8ef26171ba198/src/main/kotlin/org/wfanet/measurement/common/crypto/SignedStore.kt#L41)):*
> ```Kotlin
>     content: Flow<ByteString>
>   ): String {
>     // Since StorageClient has no concept of "overwriting" a blob, we first delete existing blobs.
> ```

We should change the StorageClient API if we actually need to support the overwrite use case.

Note that this may just be a matter of adjusting comments, as I don't think any of our existing implementations will actually stop you from overwriting a blob with the same key.
___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/SignedStore.kt` line 53 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209#-Nc8w_g87JzNLbm6y5NN:-Nc8w_g87JzNLbm6y5NO:b8uxsa0) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/766ed7840600cb1ae43ead377fd8ef26171ba198/src/main/kotlin/org/wfanet/measurement/common/crypto/SignedStore.kt#L53)):*
> ```Kotlin
>   }
> 
>   @Throws(BlobNotFoundException::class)
> ```

The `@Throws` annotation is mainly for compatibility with Java code. It is unlikely that Java code would use this library given the use of coroutines. Instead use the `@throws` KDoc tag in the function documentation.

As an aside, what's the reason for using an exception rather than returning null?

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209)* status: all files reviewed, 6 unresolved discussions (waiting on @jojijac0b and @Marco-Premier)

___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/EncryptedSignedStore.kt` line 38 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209#-Nc98anABAscGf_2vWWD:-Nc98anABAscGf_2vWWE:b-so2z55) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/766ed7840600cb1ae43ead377fd8ef26171ba198/src/main/kotlin/org/wfanet/measurement/common/crypto/EncryptedSignedStore.kt#L38)):*
> ```Kotlin
>   //  PrivateKey and X509Certificate used to create the signature when writing the data
>   //  PublicKey used to encrypt the content
>   suspend fun write(
> ```

1. The pattern we've been using with encrypted and signed content is sign-then-encrypt, meaning the signature is part of the ciphertext. Why does this use encrypt-then-sign (i.e. why are the crypto properties we're seeking different for this use case)?
2. Is it okay to have the whole blob in memory? Hybrid encryption is efficient for large messages, but you need to have the entire blob in memory to encrypt/decrypt. Tink provides a streaming implementation for AEAD, but this is a symmetric scheme meaning you may need to do your own key exchange e.g. using hybrid encryption to pass an encrypted symmetric key.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209)* status: all files reviewed, 6 unresolved discussions (waiting on @jojijac0b, @Marco-Premier, and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/EncryptedSignedStore.kt` line 38 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209#-Nc98anABAscGf_2vWWD:-Nc9EY6n9DhHTs2hgaZz:bavuf62) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/766ed7840600cb1ae43ead377fd8ef26171ba198/src/main/kotlin/org/wfanet/measurement/common/crypto/EncryptedSignedStore.kt#L38)):*

1. The pattern we've been using with encrypted and signed content is sign-then-encrypt, meaning the signature is part of the ciphertext. Why does this use encrypt-then-sign (i.e. why are the crypto properties we're seeking different for this use case)?
2. Is it okay to have the whole blob in memory? Hybrid encryption is efficient for large messages, but you need to have the entire blob in memory to encrypt/decrypt. Tink provides a streaming implementation for AEAD, but this is a symmetric scheme meaning you may need to do your own key exchange e.g. using hybrid encryption to pass an encrypted symmetric key.
</blockquote></details>

1. Panel exchange currently uses an encrypt then sign strategy for blob storage which was the pattern we were following.
2. This is a good point. Is this what you were thinking @SanjayVas  - https://developers.google.com/tink/encrypt-large-files-or-data-streams ?

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209)* status: all files reviewed, 6 unresolved discussions (waiting on @jojijac0b, @Marco-Premier, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/EncryptedSignedStore.kt` line 38 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209#-Nc98anABAscGf_2vWWD:-Nc9RzmjA4pi5f6ZC4Oa:b-2f6a83) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/766ed7840600cb1ae43ead377fd8ef26171ba198/src/main/kotlin/org/wfanet/measurement/common/crypto/EncryptedSignedStore.kt#L38)):*
> Panel exchange currently uses an encrypt then sign strategy for blob storage which was the pattern we were following.

Is this code intended to be used for panel exchange exclusively, or is this (also) for population? If for the latter, what is the cryptographic reasoning for following this pattern over sign-then-encrypt? (Just copying the way we do it for panel exchange without considering the cryptographic implications isn't a sound approach. Sign-then-encrypt is generally seen as the safer pattern, so we need to justify whenever we're not using it.)

> Is this what you were thinking [@SanjayVas](https://github.com/SanjayVas) - [https://developers.google.com/tink/encrypt-large-files-or-data-streams](https://developers.google.com/tink/encrypt-large-files-or-data-streams) ?

Yes.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209)* status: all files reviewed, 6 unresolved discussions (waiting on @jojijac0b, @Marco-Premier, and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/EncryptedSignedStore.kt` line 38 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209#-Nc98anABAscGf_2vWWD:-NcJ2Jcn3OZwObgNe4Ev:bj4l2r3) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/766ed7840600cb1ae43ead377fd8ef26171ba198/src/main/kotlin/org/wfanet/measurement/common/crypto/EncryptedSignedStore.kt#L38)):*

> Panel exchange currently uses an encrypt then sign strategy for blob storage which was the pattern we were following.

Is this code intended to be used for panel exchange exclusively, or is this (also) for population? If for the latter, what is the cryptographic reasoning for following this pattern over sign-then-encrypt? (Just copying the way we do it for panel exchange without considering the cryptographic implications isn't a sound approach. Sign-then-encrypt is generally seen as the safer pattern, so we need to justify whenever we're not using it.)

> Is this what you were thinking [@SanjayVas](https://github.com/SanjayVas) - [https://developers.google.com/tink/encrypt-large-files-or-data-streams](https://developers.google.com/tink/encrypt-large-files-or-data-streams) ?

Yes.
</blockquote></details>

I'm okay with switching it to sign-then-encrypt.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209)* status: all files reviewed, 6 unresolved discussions (waiting on @jojijac0b and @Marco-Premier)

___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/EncryptedSignedStore.kt` line 38 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209#-Nc98anABAscGf_2vWWD:-NcNJOPF7iKAB4YJdPIl:b-fciwtm) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/766ed7840600cb1ae43ead377fd8ef26171ba198/src/main/kotlin/org/wfanet/measurement/common/crypto/EncryptedSignedStore.kt#L38)):*

I'm okay with switching it to sign-then-encrypt.
</blockquote></details>

Question @SanjayVas - does this mean we would encrypt both pieces of the signed data -- the signature and the data or just the data. Feels like leaking the plaintext signature could tell you properties about the unencrypted data so safest would be to also encrypt the signature.

@jojijac0b - I suggest leaving SignedStore as-is but altering the EncryptedSignedStore to not use the underlying SignedStore but instead to follow its only flow of sign and then encrypt as Sanjay suggested. I could see adopters using either just a SignedStore or an EncryptedSignedStore so nice to give them both options.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209)* status: all files reviewed, 6 unresolved discussions (waiting on @jojijac0b and @Marco-Premier)

___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/EncryptedSignedStore.kt` line 38 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209#-Nc98anABAscGf_2vWWD:-NcNeHuh13wYSnJ2P-LU:b-yczd2b) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/766ed7840600cb1ae43ead377fd8ef26171ba198/src/main/kotlin/org/wfanet/measurement/common/crypto/EncryptedSignedStore.kt#L38)):*
> does this mean we would encrypt both pieces of the signed data

For sign-then-encrypt, you combine the content and signature into a single blob then encrypt that to get a single ciphertext. It means you need to use some format where you can put both together. e.g. `PKCS #7` or a protobuf message similar to [`SignedData`](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/main/src/main/proto/wfa/measurement/api/v2alpha/crypto.proto) from the CMMS public API.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209)* status: all files reviewed, 6 unresolved discussions (waiting on @jojijac0b and @Marco-Premier)

___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/EncryptedSignedStore.kt` line 38 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209#-Nc98anABAscGf_2vWWD:-NcYM5VQE1-mpjeH7v8B:b-rba83m) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/766ed7840600cb1ae43ead377fd8ef26171ba198/src/main/kotlin/org/wfanet/measurement/common/crypto/EncryptedSignedStore.kt#L38)):*

> does this mean we would encrypt both pieces of the signed data

For sign-then-encrypt, you combine the content and signature into a single blob then encrypt that to get a single ciphertext. It means you need to use some format where you can put both together. e.g. `PKCS #7` or a protobuf message similar to [`SignedData`](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/main/src/main/proto/wfa/measurement/api/v2alpha/crypto.proto) from the CMMS public API.
</blockquote></details>

I don't think that works well for encrypting with a stream, right? -- eg just encrypt the signature in-memory but stream encrypt the data separately -- not to the same message. The store will keep track of the two blobs itself. Does that work for you @SanjayVas ?

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209)* status: all files reviewed, 6 unresolved discussions (waiting on @jojijac0b and @Marco-Premier)

___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/EncryptedSignedStore.kt` line 38 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209#-Nc98anABAscGf_2vWWD:-NcYQspA3oxyBG4XyKnp:b-qn1loy) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/766ed7840600cb1ae43ead377fd8ef26171ba198/src/main/kotlin/org/wfanet/measurement/common/crypto/EncryptedSignedStore.kt#L38)):*
> I don't think that works well for encrypting with a stream, right

It depends on how you write the combined content + signature structure. For example consider a structure where you have the content followed by the signature, with a header/footer for each. Such a structure should allow for streaming all the way through.

> eg just encrypt the signature in-memory but stream encrypt the data separately -- not to the same message. The store will keep track of the two blobs itself.

Traditionally with sign-then-encrypt you are doing a single encryption. Doing two separate encryptions might change the security implications.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209)* status: all files reviewed, 6 unresolved discussions (waiting on @jojijac0b and @Marco-Premier)

___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/EncryptedSignedStore.kt` line 38 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209#-Nc98anABAscGf_2vWWD:-NccH88e60lrAYadfR4E:brti7kl) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/766ed7840600cb1ae43ead377fd8ef26171ba198/src/main/kotlin/org/wfanet/measurement/common/crypto/EncryptedSignedStore.kt#L38)):*

> I don't think that works well for encrypting with a stream, right

It depends on how you write the combined content + signature structure. For example consider a structure where you have the content followed by the signature, with a header/footer for each. Such a structure should allow for streaming all the way through.

> eg just encrypt the signature in-memory but stream encrypt the data separately -- not to the same message. The store will keep track of the two blobs itself.

Traditionally with sign-then-encrypt you are doing a single encryption. Doing two separate encryptions might change the security implications.
</blockquote></details>

@SanjayVas - what do you think about using https://docs.oracle.com/javase/8/docs/api/java/util/Scanner.html for this?

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209)* status: all files reviewed, 6 unresolved discussions (waiting on @jojijac0b and @Marco-Premier)

___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/EncryptedSignedStore.kt` line 38 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209#-Nc98anABAscGf_2vWWD:-NcdIUIGAe-E_PTIS_B-:b-w15kbr) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/766ed7840600cb1ae43ead377fd8ef26171ba198/src/main/kotlin/org/wfanet/measurement/common/crypto/EncryptedSignedStore.kt#L38)):*

@SanjayVas - what do you think about using https://docs.oracle.com/javase/8/docs/api/java/util/Scanner.html for this?
</blockquote></details>

I'm not sure I understand where that would fit into the implementation. The details don't matter as much so long as we're doing things chunked rather than byte-by-byte. e.g. using `ByteChannel` instead of `InputStream`/`OutputStream`. Note that the `StreamingAead` primitive has `ByteChannel` methods. We already have functions in common-jvm to interface between coroutine `Flow` and `ByteChannel`, as the GCS Java API is also `ByteChannel`-based.

stevenwarejones (approved):
> Reviewed 6 of 6 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209)* status: all files reviewed, 6 unresolved discussions (waiting on @jojijac0b and @Marco-Premier)

SanjayVas (commented):
From offline discussion, I still don't see a use case for opaquely storing the content and signature as two separate blobs.

1. If we're going to need this to be transparent (e.g. callers need to be able to interact with storage directly without necessarily using this code), then we shouldn't be hiding the use of two blobs as an implementation detail.
2. If we're going to want to encrypt the content as well, it should be done with sign-then-encrypt with a single ciphertext. In that case, the plaintext content and signature should never be stored.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209#-:-Ne4b8iR0ClRpnazT7AF:btv5psm)* status: all files reviewed, 6 unresolved discussions (waiting on @jojijac0b and @Marco-Premier)

stevenwarejones (approved):
@SanjayVas - I believe the design here was across two use cases. I agree with you on the key transparency need that probably negates using the Store interface for VID model blobs. However, the other use case we were working toward was replacing the existing VerifyingStorageClient and SigningStorageClient used in panel exchange with something based around Store since the signatures are implementation details in that case. lmk what you think.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209#-:-Nenxr3w4kAciePUEwvr:b-4liytd)* status: all files reviewed, 6 unresolved discussions (waiting on @jojijac0b and @Marco-Premier)

SanjayVas (commented):
> replacing the existing VerifyingStorageClient and SigningStorageClient used in panel exchange with something based around Store since the signatures are implementation details in that case. lmk what you think.

Perhaps that should stay in the panel match tree in the main repo then, since I don't think it's a pattern that we necessarily want to replicate elsewhere.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/209#-:-NessVOX2ru4OIkaPY2t:bb500s9)* status: all files reviewed, 6 unresolved discussions (waiting on @jojijac0b and @Marco-Premier)

PR #218, opened 2023-09-28
Specify signature algorithm in signature utils.

### LINKED ISSUE #217: Signatures.kt uses signature algorithm rather than public key algorithm with digest
Reported by @stevenwarejones.

[Signatures.kt](https://github.com/world-federation-of-advertisers/common-jvm/blob/ec2709d2412be038c513c995cb7ad1a5e4aaa201/src/main/kotlin/org/wfanet/measurement/common/crypto/Signatures.kt) uses the signature algorithm from the X.509 certificate for signing. This is actually the algorithm for the signature within the certificate, signed using the issuer's private key. We should instead be using the algorithm associated with the public key.

Our code basically assumes that the public key algorithm of the certificate is the same as the one used by the issuer. This is commonly the case, which is why we had not seen any errors.

Note that the public key algorithm is technically just the encryption algorithm, whereas a signature algorithm also includes a digest algorithm. For example, the public key algorithm may be `id-ecPublicKey` which means the key can be used for `ecdsa-with-SHA256`. This also means that our API may be underspecified when it comes to digital signature types.

For the Java security library, we need to use the [Java security standard algorithm names](https://docs.oracle.com/en/java/javase/11/docs/specs/security/standard-names.html).

SanjayVas:
Closes #217


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/218)

stevenwarejones (changes requested):
> Reviewed 12 of 12 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/218)* status: all files reviewed, 5 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/Signatures.kt` line 56 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/218#-NfWKClTDsGQf8Tdmqvq:-NfWKClTDsGQf8Tdmqvr:b-tte8t4) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/136c829fddddbc4396e3ac25b3d7fa72dd485a1d/src/main/kotlin/org/wfanet/measurement/common/crypto/Signatures.kt#L56)):*
> ```Kotlin
>   ECDSA_WITH_SHA384("1.2.840.10045.4.3.3", "SHA384withECDSA"),
>   ECDSA_WITH_SHA512("1.2.840.10045.4.3.4", "SHA512withECDSA"),
>   DSA_WITH_SHA256("2.16.840.1.101.3.4.3.2", "SHA256withDSA"),
> ```

see other comment about not supporting DSA
___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/Signatures.kt` line 93 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/218#-NfWK4f25D5T-bV1ohpt:-NfWK4f25D5T-bV1ohpu:b-bafxf5) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/136c829fddddbc4396e3ac25b3d7fa72dd485a1d/src/main/kotlin/org/wfanet/measurement/common/crypto/Signatures.kt#L93)):*
> ```Kotlin
>             HashAlgorithm.SHA512 -> SHA_512_WITH_RSA_ENCRYPTION
>           }
>         "DSA" ->
> ```

i don't think this is supported by conscrypt so we probably don't need to support here
___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/Signatures.kt` line 99 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/218#-NfWJsD_7Y2Msgw0VWmI:-NfWJsD_7Y2Msgw0VWmJ:b-dwjwas) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/136c829fddddbc4396e3ac25b3d7fa72dd485a1d/src/main/kotlin/org/wfanet/measurement/common/crypto/Signatures.kt#L99)):*
> ```Kotlin
>             HashAlgorithm.SHA512 -> null
>           }
>         else -> null
> ```

what about RSA/PSS ? including adding separate tests for both `withRSA` and `withRSA/PSS`
___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/Signatures.kt` line 138 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/218#-NfWM6M_1ycW5_1FCAVw:-NfWM6M_1ycW5_1FCAVx:b-dt0tb3) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/136c829fddddbc4396e3ac25b3d7fa72dd485a1d/src/main/kotlin/org/wfanet/measurement/common/crypto/Signatures.kt#L138)):*
> ```Kotlin
>  * @throws java.security.NoSuchAlgorithmException if no implementation is found for the specified
>  *   algorithm in [jceProvider]
>  * @throws java.security.InvalidKeyException if the [PrivateKey] is invalid, including if it is not
> ```

or if the algorithm specified is not compatible with the PrivateKey
___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/SignedBlob.kt` line 60 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/218#-NfWLLpa0eg_T8GEUNdf:-NfWLLpa0eg_T8GEUNdg:b-9pno91) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/136c829fddddbc4396e3ac25b3d7fa72dd485a1d/src/main/kotlin/org/wfanet/measurement/common/crypto/SignedBlob.kt#L60)):*
> ```Kotlin
>   algorithm: SignatureAlgorithm,
> ): SignedBlob {
>   val signer = privateKey.newSigner(algorithm)
> ```

why not just call the other createSignedBlob
```
val signer = privateKey.newSigner(algorithm)
return StorageClient.createSignedBlob(blobKey, content, signer)
```

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/218)* status: all files reviewed, 5 unresolved discussions (waiting on @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/Signatures.kt` line 56 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/218#-NfWKClTDsGQf8Tdmqvq:-NfXTHSZ4ETzAtDARdTl:b-cx786g) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/136c829fddddbc4396e3ac25b3d7fa72dd485a1d/src/main/kotlin/org/wfanet/measurement/common/crypto/Signatures.kt#L56)):*

see other comment about not supporting DSA
</blockquote></details>

Done (dropped)
___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/Signatures.kt` line 93 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/218#-NfWK4f25D5T-bV1ohpt:-NfXTHpT3n3R8eLFG_R8:b-cx786g) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/136c829fddddbc4396e3ac25b3d7fa72dd485a1d/src/main/kotlin/org/wfanet/measurement/common/crypto/Signatures.kt#L93)):*

i don't think this is supported by conscrypt so we probably don't need to support here
</blockquote></details>

Done (dropped)
___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/Signatures.kt` line 99 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/218#-NfWJsD_7Y2Msgw0VWmI:-NfXTJMoBB69-WXOuvlB:bo6ycwa) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/136c829fddddbc4396e3ac25b3d7fa72dd485a1d/src/main/kotlin/org/wfanet/measurement/common/crypto/Signatures.kt#L99)):*

what about RSA/PSS ? including adding separate tests for both `withRSA` and `withRSA/PSS`
</blockquote></details>

RSA-PSS is intentionally not included because it's more complicated to support. It requires additional parameters beyond the key and algorithm identifier to be configured in the signer/verifier.

Note that it's totally fine if the certificate algorithm is RSA-PSS (or more specifically, RSASSA-PSS) as long as the certificate has a regular RSA or EC key. (You can use either a regular RSA key for RSA-PSS, or you can use the specific RSA-PSS key type. We only support the former).
___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/Signatures.kt` line 138 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/218#-NfWM6M_1ycW5_1FCAVw:-NfXTPRkDBOtB3S5CZVj:bg1i3pa) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/136c829fddddbc4396e3ac25b3d7fa72dd485a1d/src/main/kotlin/org/wfanet/measurement/common/crypto/Signatures.kt#L138)):*

or if the algorithm specified is not compatible with the PrivateKey
</blockquote></details>

That is what the comment already says, or are you suggesting a slight rephrasing?
___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/SignedBlob.kt` line 60 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/218#-NfWLLpa0eg_T8GEUNdf:-NfXUpM003gVN_bRHjqE:bota9mz) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/136c829fddddbc4396e3ac25b3d7fa72dd485a1d/src/main/kotlin/org/wfanet/measurement/common/crypto/SignedBlob.kt#L60)):*

why not just call the other createSignedBlob
```
val signer = privateKey.newSigner(algorithm)
return StorageClient.createSignedBlob(blobKey, content, signer)
```

</blockquote></details>

That one derives the signing algorithm from the signer rather than using the one from the argument. This overload is specifically for the case where you know the algorithm and don't already have a signer factory.

stevenwarejones (changes requested):
> Reviewed 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/218)* status: all files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/Signatures.kt` line 138 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/218#-NfWM6M_1ycW5_1FCAVw:-NfXc48uAM_coZGx9t-F:b6r33s7) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/136c829fddddbc4396e3ac25b3d7fa72dd485a1d/src/main/kotlin/org/wfanet/measurement/common/crypto/Signatures.kt#L138)):*

That is what the comment already says, or are you suggesting a slight rephrasing?
</blockquote></details>

yeah - I meant a slight rephrase

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/218)* status: 11 of 12 files reviewed, 1 unresolved discussion (waiting on @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/Signatures.kt` line 138 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/218#-NfWM6M_1ycW5_1FCAVw:-NfXnG2g0c1c-0Y7Q-vI:b-so1me5) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/136c829fddddbc4396e3ac25b3d7fa72dd485a1d/src/main/kotlin/org/wfanet/measurement/common/crypto/Signatures.kt#L138)):*

yeah - I meant a slight rephrase
</blockquote></details>

Rephrased slightly, though not using the suggested phrasing. Since it's an InvalidKeyException, the condition is that the key is invalid (for this function call). This can mean that it's incompatible with the specified algorithm. It could also just be invalid regardless of the algorithm, but I suppose that's just a subset of being invalid for the specified algorithm.

stevenwarejones (changes requested):
> Reviewed 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/218)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/Signatures.kt` line 138 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/218#-NfWM6M_1ycW5_1FCAVw:-NfkEXIU3YxGP-W5yt1N:bob6hyg) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/136c829fddddbc4396e3ac25b3d7fa72dd485a1d/src/main/kotlin/org/wfanet/measurement/common/crypto/Signatures.kt#L138)):*

Rephrased slightly, though not using the suggested phrasing. Since it's an InvalidKeyException, the condition is that the key is invalid (for this function call). This can mean that it's incompatible with the specified algorithm. It could also just be invalid regardless of the algorithm, but I suppose that's just a subset of being invalid for the specified algorithm.
</blockquote></details>

perfect
___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/SignedBlob.kt` line 60 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/218#-NfWLLpa0eg_T8GEUNdf:-NfkEb7GATYhIrrcwbh7:bdwkzi6) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/136c829fddddbc4396e3ac25b3d7fa72dd485a1d/src/main/kotlin/org/wfanet/measurement/common/crypto/SignedBlob.kt#L60)):*

That one derives the signing algorithm from the signer rather than using the one from the argument. This overload is specifically for the case where you know the algorithm and don't already have a signer factory.
</blockquote></details>

i'm just saying this you are creating a signer here anyways, you can just use the other function to share code

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/218)* status: all files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/218)* status: all files reviewed, 2 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/SigningKeyHandle.kt` line 29 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/218#-NfkFxv-4K58Ysw6AKN3:-NfkFxv04CGYmgxwy5gW:b-xt981d) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/7c8ede2e9eedd0cafe8ac6b1335d33607462762b/src/main/kotlin/org/wfanet/measurement/common/crypto/SigningKeyHandle.kt#L29)):*
> ```Kotlin
>   /** Default [SignatureAlgorithm] for [privateKey] using [DEFAULT_HASH_ALGORITHM]. */
>   val defaultAlgorithm =
>     requireNotNull(SignatureAlgorithm.fromKeyAndHashAlgorithm(privateKey, DEFAULT_HASH_ALGORITHM)) {
> ```

this should be `by lazy`. It may not be needed. Its also possible that someone may what to use something that will throw an error here.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/218)* status: all files reviewed, 2 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/SigningKeyHandle.kt` line 29 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/218#-NfkFxv-4K58Ysw6AKN3:-NfkHM0v97huzo85zWUN:b5jq0vn) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/7c8ede2e9eedd0cafe8ac6b1335d33607462762b/src/main/kotlin/org/wfanet/measurement/common/crypto/SigningKeyHandle.kt#L29)):*

this should be `by lazy`. It may not be needed. Its also possible that someone may what to use something that will throw an error here.
</blockquote></details>

^want to use something

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/218)* status: 11 of 12 files reviewed, 2 unresolved discussions (waiting on @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/SignedBlob.kt` line 60 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/218#-NfWLLpa0eg_T8GEUNdf:-Nfl81qC4AUPMOnAjzNd:b5jdoog) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/136c829fddddbc4396e3ac25b3d7fa72dd485a1d/src/main/kotlin/org/wfanet/measurement/common/crypto/SignedBlob.kt#L60)):*

i'm just saying this you are creating a signer here anyways, you can just use the other function to share code
</blockquote></details>

Calling the other overload would result in an unnecessary re-conversion from the string Java algorithm name to the `SignatureAlgorithm` rather than using the `SignatureAlgorithm` that is already available.
___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/SigningKeyHandle.kt` line 29 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/218#-NfkFxv-4K58Ysw6AKN3:-Nfl8btl6ZYX_mhPwYt6:b-gw8wi1) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/7c8ede2e9eedd0cafe8ac6b1335d33607462762b/src/main/kotlin/org/wfanet/measurement/common/crypto/SigningKeyHandle.kt#L29)):*
> Its also possible that someone may want to use something that will throw an error here.

That is intentionally not supported. The default algorithm should always be valid.

> this should be `by lazy`. It may not be needed.

It's cheap to initialize, and failing fast is desired here if the private key does not have a `SignatureAlgorithm` that is compatible with `DEFAULT_HASH_ALGORITHM`. I updated the constructor documentation to make this clearer.

If calling a property accessor result in an exception (usually `IllegalStateException`), it indicates a programming error. This is why `DEFAULT_HASH_ALGORITHM` is public, as it allows the caller to avoid the exception.

stevenwarejones (approved):
> Reviewed 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/218)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @SanjayVas)

PR #223, opened 2023-11-14
Include OpenTelemetry trace ID in gRPC logs when available.

SanjayVas:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/223)

PR #227, opened 2023-12-13
Include runfiles in java_image.

SanjayVas:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/227)

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/227)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`build/rules_oci/defs.bzl` line 108 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/227#-NlcyNFqBzI2j8jLqmUN:-NlcyNFqBzI2j8jLqmUO:b7tfbd5) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/027415d85994e389160d5ea7ed7d19576be5d21f/build/rules_oci/defs.bzl#L108)):*
> ```Starlark
>       **kwargs: other args to pass to the resulting target
>     """
>     env = env or {}
> ```

why not just set the default argument to `{}`?

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/227)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @stevenwarejones)

___
*[`build/rules_oci/defs.bzl` line 108 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/227#-NlcyNFqBzI2j8jLqmUN:-NldOyhHEzFXL4OISox2:buuabjp) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/027415d85994e389160d5ea7ed7d19576be5d21f/build/rules_oci/defs.bzl#L108)):*
From https://bazel.build/extending/macros:

> In most cases, optional parameters should have a default value of `None`. `None` can be passed directly to native rules, which treat it the same as if you had not passed in any argument. Thus, there is no need to replace it with `0`, `False`, or `[]` for this purpose. Instead, the macro should defer to the rules it creates, as their defaults may be complex or may change over time. Additionally, a parameter that is explicitly set to its default value looks different than one that is never set (or set to `None`) when accessed through the query language or build-system internals.

SanjayVas (commented):
> Dismissed @stevenwarejones from a discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/227)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @stevenwarejones)

SanjayVas (commented):
> Reviewed 1 of 1 files at r1.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/227)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @stevenwarejones)

PR #242, opened 2024-04-02
Add function getAead for TinkPrivateKeyHandle.

renjiezh:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/242)

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/242)* status: all files reviewed, 1 unresolved discussion (waiting on @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/tink/KeyHandle.kt` line 74 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/242#-NuVhaJd6Dtr9vOXsAxU:-NuVhaJd6Dtr9vOXsAxV:b-3bcwaq) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/ce4854463239798e79f92321cf2b32cd2fbf7b5e/src/main/kotlin/org/wfanet/measurement/common/crypto/tink/KeyHandle.kt#L74)):*
> ```Kotlin
>   }
> 
>   fun getAead(): Aead {
> ```

This doesn't make sense here. TinkPrivateKeyHandle is for asymmetric keys.

PR #252, opened 2024-06-05
Upgrade AWS SDK version to 2.23.18

sandip80:
## Changes
* Upgrade AWS SDK version to bring parity with EMR image version 7.1.0 used for panel exchange step execution.

## Testing
* Unit test
* E2E test (on AWS cloud)


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/252)

sandip80:
*[`src/main/kotlin/org/wfanet/measurement/aws/s3/S3StorageClient.kt` line 164 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/252#-NzdbouFDRxjLQtmrKzJ:-Nze18qiBnQSneHgMcUd:bdrshv0) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/c30c552f78ac0fb0a33b8b635c44d51e2ff23ce3/src/main/kotlin/org/wfanet/measurement/aws/s3/S3StorageClient.kt#L164)):*

Can you clarify why this change is needed? Comment says something about conflicting load for Spark. Is that implying Spark is somehow not loading the same SDK version?

Doing it this way means we're spinning up a background thread to do the blocking read, where we really want to avoid blocking reads wherever possible.
</blockquote></details>

Yes, so within EMR spark image, it comes with preinstalled AWS SDK which is also included in the classpath for the spark app. Java will try loading this classpath pre-included in the spark-submit param. When the spark app runs and tries to use the already loaded classes from the classpath, this causes ClassCastException (when the AsyncResponseTransformer tries to return the Publisher as a ResponsePublisher which implements the reactivestream Publisher) here.

sandip80:
*[`src/main/kotlin/org/wfanet/measurement/aws/s3/S3StorageClient.kt` line 164 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/252#-NzdbouFDRxjLQtmrKzJ:-Nze3tV68teJPrZ1HlnM:b-7gjs45) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/c30c552f78ac0fb0a33b8b635c44d51e2ff23ce3/src/main/kotlin/org/wfanet/measurement/aws/s3/S3StorageClient.kt#L164)):*
<details><summary><i>Previously, sandip80 (Sandip Samantaray) wrote…</i></summary><blockquote>

Yes, so within EMR spark image, it comes with preinstalled AWS SDK which is also included in the classpath for the spark app. Java will try loading this classpath pre-included in the spark-submit param. When the spark app runs and tries to use the already loaded classes from the classpath, this causes ClassCastException (when the AsyncResponseTransformer tries to return the Publisher as a ResponsePublisher which implements the reactivestream Publisher) here.
</blockquote></details>

Forcing the Spark app to load the user classpath first also leads to other exceptions related to AWS Client interceptors.

stevenwarejones (approved):
> Reviewed 3 of 4 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/252)* status: all files reviewed, 2 unresolved discussions (waiting on @sandip80 and @SanjayVas)

___
*[`src/test/kotlin/org/wfanet/measurement/aws/s3/S3StorageClientTest.kt` line 56 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/252#-Nze7YDM-Jjy4hKauhMc:-Nze7YDM-Jjy4hKauhMd:bthl1ui) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/c30c552f78ac0fb0a33b8b635c44d51e2ff23ce3/src/test/kotlin/org/wfanet/measurement/aws/s3/S3StorageClientTest.kt#L56)):*
> ```Kotlin
>     .region(Region.of("us-east-1"))
>     .credentialsProvider(StaticCredentialsProvider.create(AwsBasicCredentials.create("foo", "bar")))
>     .serviceConfiguration(S3Configuration.builder().pathStyleAccessEnabled(true).build())
> ```

interesting. you needed this to get the test to pass? does the storageFactory need updating? https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/14f0813ee06e668e6d74633961c48da89c2febbb/src/main/kotlin/org/wfanet/panelmatch/client/storage/aws/s3/S3StorageFactory.kt#L34

sandip80:
*[`src/test/kotlin/org/wfanet/measurement/aws/s3/S3StorageClientTest.kt` line 56 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/252#-Nze7YDM-Jjy4hKauhMc:-NzeCzW609GCoB1oHos5:b-tjbri1) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/c30c552f78ac0fb0a33b8b635c44d51e2ff23ce3/src/test/kotlin/org/wfanet/measurement/aws/s3/S3StorageClientTest.kt#L56)):*

interesting. you needed this to get the test to pass? does the storageFactory need updating? https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/14f0813ee06e668e6d74633961c48da89c2febbb/src/main/kotlin/org/wfanet/panelmatch/client/storage/aws/s3/S3StorageFactory.kt#L34
</blockquote></details>

Yeah, this is in the createS3ClientV2 code of the mock rule. However, in production this isn't recommended since AWS has been planning to deprecate path style access in favor of virtual host access: https://aws.amazon.com/blogs/aws/amazon-s3-path-deprecation-plan-the-rest-of-the-story/

I think we will need to update this code since S3MockRule has been marked for deprecation (https://github.com/adobe/S3Mock/blob/main/testsupport/junit4/src/main/java/com/adobe/testing/s3mock/junit4/S3MockRule.java#L42C2-L44C4) in favor of Junit5 based Extension.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/252)* status: all files reviewed, 2 unresolved discussions (waiting on @sandip80 and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/aws/s3/S3StorageClient.kt` line 164 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/252#-NzdbouFDRxjLQtmrKzJ:-NzeWWOS4CuL3jHweFqJ:bxw1uc1) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/c30c552f78ac0fb0a33b8b635c44d51e2ff23ce3/src/main/kotlin/org/wfanet/measurement/aws/s3/S3StorageClient.kt#L164)):*
<details><summary><i>Previously, sandip80 (Sandip Samantaray) wrote…</i></summary><blockquote>

Forcing the Spark app to load the user classpath first also leads to other exceptions related to AWS Client interceptors.
</blockquote></details>

Doesn't this mean that we need to keep our AWS SDK at the same version as what's in the Spark image? Or is the issue that the Spark image is missing some part of the SDK so we need to limit ourselves to the same parts?

Whatever it is, the restrictions should be documented here.

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/252)* status: all files reviewed, 2 unresolved discussions (waiting on @sandip80)

___
*[`src/test/kotlin/org/wfanet/measurement/aws/s3/S3StorageClientTest.kt` line 56 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/252#-Nze7YDM-Jjy4hKauhMc:-NzhbL_Q-HqJuXjf0GxQ:b-3a0uqr) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/c30c552f78ac0fb0a33b8b635c44d51e2ff23ce3/src/test/kotlin/org/wfanet/measurement/aws/s3/S3StorageClientTest.kt#L56)):*
<details><summary><i>Previously, sandip80 (Sandip Samantaray) wrote…</i></summary><blockquote>

Yeah, this is in the createS3ClientV2 code of the mock rule. However, in production this isn't recommended since AWS has been planning to deprecate path style access in favor of virtual host access: https://aws.amazon.com/blogs/aws/amazon-s3-path-deprecation-plan-the-rest-of-the-story/

I think we will need to update this code since S3MockRule has been marked for deprecation (https://github.com/adobe/S3Mock/blob/main/testsupport/junit4/src/main/java/com/adobe/testing/s3mock/junit4/S3MockRule.java#L42C2-L44C4) in favor of Junit5 based Extension.
</blockquote></details>

can you add a TODO to get rid of the S3MockRule?

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/252)* status: all files reviewed, 2 unresolved discussions (waiting on @sandip80)

___
*[`src/main/kotlin/org/wfanet/measurement/aws/s3/S3StorageClient.kt` line 164 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/252#-NzdbouFDRxjLQtmrKzJ:-Nzjx8x1ApkaPmo6QpUZ:b2mdvf7) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/c30c552f78ac0fb0a33b8b635c44d51e2ff23ce3/src/main/kotlin/org/wfanet/measurement/aws/s3/S3StorageClient.kt#L164)):*

Doesn't this mean that we need to keep our AWS SDK at the same version as what's in the Spark image? Or is the issue that the Spark image is missing some part of the SDK so we need to limit ourselves to the same parts?

Whatever it is, the restrictions should be documented here.
</blockquote></details>

Thinking about this more, I really think this is a bad idea. We're basically acknowledging that we might have a different SDK version under some conditions. This code change here is just one example of working around a version mismatch - there may be others that aren't so obvious.

We need to have a way to ensure that the SDK version used in EMR Spark matches the one we pull in from Maven. This could either be figuring out how to get the Spark image to not load its own SDK or documenting what SDK version Spark uses and depending on that version in common-jvm.

If there is *really* no way to ensure we have the same versions, then the last resort would be to create a separate S3 StorageClient impl just for EMR Spark that lives in the main repo.

sandip80:
*[`src/main/kotlin/org/wfanet/measurement/aws/s3/S3StorageClient.kt` line 164 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/252#-NzdbouFDRxjLQtmrKzJ:-O-2sbgg2jWGVm2MuU4M:bl5it6r) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/c30c552f78ac0fb0a33b8b635c44d51e2ff23ce3/src/main/kotlin/org/wfanet/measurement/aws/s3/S3StorageClient.kt#L164)):*

Thinking about this more, I really think this is a bad idea. We're basically acknowledging that we might have a different SDK version under some conditions. This code change here is just one example of working around a version mismatch - there may be others that aren't so obvious.

We need to have a way to ensure that the SDK version used in EMR Spark matches the one we pull in from Maven. This could either be figuring out how to get the Spark image to not load its own SDK or documenting what SDK version Spark uses and depending on that version in common-jvm.

If there is *really* no way to ensure we have the same versions, then the last resort would be to create a separate S3 StorageClient impl just for EMR Spark that lives in the main repo.
</blockquote></details>

Having same SDK version as the EMR image doesn't resolve this issue. The issue is due to a different classloader loading the Publisher interface other than the app classloader. And when the logic here tries to cast to ResponsePublisher to Publisher it fails. There is no way to solve this as of now. I will create a separate S3 StorageClient impl for EMR spark.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/252)* status: 2 of 4 files reviewed, 2 unresolved discussions (waiting on @sandip80 and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/aws/s3/S3StorageClient.kt` line 164 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/252#-NzdbouFDRxjLQtmrKzJ:-O-3A1q02q-fxCMjlnWM:b-ezsyc4) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/c30c552f78ac0fb0a33b8b635c44d51e2ff23ce3/src/main/kotlin/org/wfanet/measurement/aws/s3/S3StorageClient.kt#L164)):*
> Having same SDK version as the EMR image doesn't resolve this issue. The issue is due to a different classloader loading the Publisher interface other than the app classloader.

Having the same SDK in the EMR image means having the same SDK classes loaded. If different classes are being loaded, that by definition means that it's using a different SDK version.

sandip80:
*[`src/main/kotlin/org/wfanet/measurement/aws/s3/S3StorageClient.kt` line 164 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/252#-NzdbouFDRxjLQtmrKzJ:-O-40BDO2eEgN7eRpwye:b9fdg5l) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/c30c552f78ac0fb0a33b8b635c44d51e2ff23ce3/src/main/kotlin/org/wfanet/measurement/aws/s3/S3StorageClient.kt#L164)):*

> Having same SDK version as the EMR image doesn't resolve this issue. The issue is due to a different classloader loading the Publisher interface other than the app classloader.

Having the same SDK in the EMR image means having the same SDK classes loaded. If different classes are being loaded, that by definition means that it's using a different SDK version.
</blockquote></details>

I am not sure there is a way to ensure the versions are the same for Reactive Stream. I am following AWS documentation for EMR image to determine the AWS SDK version: https://docs.aws.amazon.com/emr/latest/ReleaseGuide/emr-710-release.html

sandip80:
*[`src/main/kotlin/org/wfanet/measurement/aws/s3/S3StorageClient.kt` line 164 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/252#-NzdbouFDRxjLQtmrKzJ:-O-41Emt6WSeM6bWaAsT:b-inw6z0) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/c30c552f78ac0fb0a33b8b635c44d51e2ff23ce3/src/main/kotlin/org/wfanet/measurement/aws/s3/S3StorageClient.kt#L164)):*
<details><summary><i>Previously, sandip80 (Sandip Samantaray) wrote…</i></summary><blockquote>

I am not sure there is a way to ensure the versions are the same for Reactive Stream. I am following AWS documentation for EMR image to determine the AWS SDK version: https://docs.aws.amazon.com/emr/latest/ReleaseGuide/emr-710-release.html
</blockquote></details>

EMR 7.1.0 uses AWS SDK 2.23.18. This PR is upgrading to the same version. But at runtime we still see the issue.

stevenwarejones (approved):
> Reviewed 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/252)* status: all files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/aws/s3/S3StorageClient.kt` line 164 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/252#-NzdbouFDRxjLQtmrKzJ:-O-6XCPY30nmhgiupho6:b-wnip3g) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/c30c552f78ac0fb0a33b8b635c44d51e2ff23ce3/src/main/kotlin/org/wfanet/measurement/aws/s3/S3StorageClient.kt#L164)):*
<details><summary><i>Previously, sandip80 (Sandip Samantaray) wrote…</i></summary><blockquote>

EMR 7.1.0 uses AWS SDK 2.23.18. This PR is upgrading to the same version. But at runtime we still see the issue.
</blockquote></details>

I believe we are able to set the EMR version and the AWS SDK version. We just need to make sure they align. Can you update comments on both deps saying they should be aligned?

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/252)* status: all files reviewed, 1 unresolved discussion (waiting on @sandip80 and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/aws/s3/S3StorageClient.kt` line 164 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/252#-NzdbouFDRxjLQtmrKzJ:-O-7v9zV3wY9cUg9W8PV:b7g5hff) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/c30c552f78ac0fb0a33b8b635c44d51e2ff23ce3/src/main/kotlin/org/wfanet/measurement/aws/s3/S3StorageClient.kt#L164)):*
> I am following AWS documentation for EMR image to determine the AWS SDK version: https://docs.aws.amazon.com/emr/latest/ReleaseGuide/emr-710-release.html

There are two SDK versions listed in that: 2.23.18 and 1.12.656. How do we control which is loaded? Or are they under different packages?

Could it be the issue isn't the AWS SDK version but the corresponding ReactiveStreams version? Looks like software.amazon.awssdk:s3:2.23.18 depends on org.reactivestreams:reactive-streams:1.0.4, and that matches what's in maven_install.json in the main repo. What version of org.reactivestreams:reactive-streams is in that image classpath?

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/252)* status: all files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/aws/s3/S3StorageClient.kt` line 164 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/252#-NzdbouFDRxjLQtmrKzJ:-O-M9KSa20hJcF67Hv9_:b-hornix) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/c30c552f78ac0fb0a33b8b635c44d51e2ff23ce3/src/main/kotlin/org/wfanet/measurement/aws/s3/S3StorageClient.kt#L164)):*

> I am following AWS documentation for EMR image to determine the AWS SDK version: https://docs.aws.amazon.com/emr/latest/ReleaseGuide/emr-710-release.html

There are two SDK versions listed in that: 2.23.18 and 1.12.656. How do we control which is loaded? Or are they under different packages?

Could it be the issue isn't the AWS SDK version but the corresponding ReactiveStreams version? Looks like software.amazon.awssdk:s3:2.23.18 depends on org.reactivestreams:reactive-streams:1.0.4, and that matches what's in maven_install.json in the main repo. What version of org.reactivestreams:reactive-streams is in that image classpath?
</blockquote></details>

looks like you can use them side by side: https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/migration-whats-different.html

sandip80:
*[`src/main/kotlin/org/wfanet/measurement/aws/s3/S3StorageClient.kt` line 164 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/252#-NzdbouFDRxjLQtmrKzJ:-O-7ezB_1DCv6V1Do1ML:bmwc4sn) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/c30c552f78ac0fb0a33b8b635c44d51e2ff23ce3/src/main/kotlin/org/wfanet/measurement/aws/s3/S3StorageClient.kt#L164)):*

I believe we are able to set the EMR version and the AWS SDK version. We just need to make sure they align. Can you update comments on both deps saying they should be aligned?
</blockquote></details>

Adding the comment for AWS SDK here. Will add EMR comment on the cmm repo.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/252)* status: 3 of 4 files reviewed, 1 unresolved discussion (waiting on @sandip80 and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/aws/s3/S3StorageClient.kt` line 164 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/252#-NzdbouFDRxjLQtmrKzJ:-O-bXeD9AXr1iFsqIJOY:b1b29an) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/c30c552f78ac0fb0a33b8b635c44d51e2ff23ce3/src/main/kotlin/org/wfanet/measurement/aws/s3/S3StorageClient.kt#L164)):*
<details><summary><i>Previously, sandip80 (Sandip Samantaray) wrote…</i></summary><blockquote>

AWS SDK version v2.x.x and v1.x.x are v2 (https://github.com/aws/aws-sdk-java-v2) and v1 (https://github.com/aws/aws-sdk-java) versions of the AWS SDK. They are both loaded together in the EMR image. The ReactiveStreams version is same as CMM repo version (1.0.4) but we continue to see the issue with the same version.
</blockquote></details>

If those are in fact the versions in the classpath used in the image, then I wouldn't expect to be seeing this issue. It still seems like there's some other version being loaded.

Linking the screenshot of the exception for context:

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/252)* status: 3 of 4 files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/aws/s3/S3StorageClient.kt` line 164 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/252#-NzdbouFDRxjLQtmrKzJ:-O-gYyEB9J_ZgRLWmoit:b-zaf8qk) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/c30c552f78ac0fb0a33b8b635c44d51e2ff23ce3/src/main/kotlin/org/wfanet/measurement/aws/s3/S3StorageClient.kt#L164)):*

If those are in fact the versions in the classpath used in the image, then I wouldn't expect to be seeing this issue. It still seems like there's some other version being loaded.

Linking the screenshot of the exception for context:


</blockquote></details>

I thought the issue had to do with the versions we have in common-jvm/cross-media-measurement not aligning with the versions in the EMR container

sandip80:
*[`src/main/kotlin/org/wfanet/measurement/aws/s3/S3StorageClient.kt` line 164 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/252#-NzdbouFDRxjLQtmrKzJ:-O-heht48EIqJ05uzamL:blh7i1f) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/c30c552f78ac0fb0a33b8b635c44d51e2ff23ce3/src/main/kotlin/org/wfanet/measurement/aws/s3/S3StorageClient.kt#L164)):*

I thought the issue had to do with the versions we have in common-jvm/cross-media-measurement not aligning with the versions in the EMR container
</blockquote></details>

Nope, we see this issue even though we have an exact match of the EMR AWS SDK version. I only saw this issue resolved in EMR version 6.14.0 which is missing AWS SDK v2 and only has v1 installed. But we cannot use older EMR version since it has an older glibc version not compatible with the cmm recommended version of glibc.

sandip80:
*[`src/main/kotlin/org/wfanet/measurement/aws/s3/S3StorageClient.kt` line 164 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/252#-NzdbouFDRxjLQtmrKzJ:-O-rEbVV5gwHcRJ7xbLg:bxh0z27) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/c30c552f78ac0fb0a33b8b635c44d51e2ff23ce3/src/main/kotlin/org/wfanet/measurement/aws/s3/S3StorageClient.kt#L164)):*
<details><summary><i>Previously, sandip80 (Sandip Samantaray) wrote…</i></summary><blockquote>

Nope, we see this issue even though we have an exact match of the EMR AWS SDK version. I only saw this issue resolved in EMR version 6.14.0 which is missing AWS SDK v2 and only has v1 installed. But we cannot use older EMR version since it has an older glibc version not compatible with the cmm recommended version of glibc.
</blockquote></details>

@stevenwarejones @SanjayVas are we good to merge this change?

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/252)* status: 3 of 4 files reviewed, 1 unresolved discussion (waiting on @sandip80 and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/aws/s3/S3StorageClient.kt` line 164 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/252#-NzdbouFDRxjLQtmrKzJ:-O0FSTTx72nPeVHiRIQX:b-ueqrtn) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/c30c552f78ac0fb0a33b8b635c44d51e2ff23ce3/src/main/kotlin/org/wfanet/measurement/aws/s3/S3StorageClient.kt#L164)):*
<details><summary><i>Previously, sandip80 (Sandip Samantaray) wrote…</i></summary><blockquote>

@stevenwarejones @SanjayVas are we good to merge this change?
</blockquote></details>

Not until the ResponsePublisher change is reverted, meaning until we can figure out how to get the loaded classes to match up between local builds and what's loaded in EMR.

My current guess is that there's another copy of the SDK somewhere in the image that's being loaded. There may need to be classpath adjustments to ensure that all the right libraries are loaded.

stevenwarejones (approved):
> Reviewed 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/252)* status: all files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/aws/s3/S3StorageClient.kt` line 164 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/252#-NzdbouFDRxjLQtmrKzJ:-O0Q-HfYAbc1Dyv01tJF:b11q4pj) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/c30c552f78ac0fb0a33b8b635c44d51e2ff23ce3/src/main/kotlin/org/wfanet/measurement/aws/s3/S3StorageClient.kt#L164)):*

Not until the ResponsePublisher change is reverted, meaning until we can figure out how to get the loaded classes to match up between local builds and what's loaded in EMR.

My current guess is that there's another copy of the SDK somewhere in the image that's being loaded. There may need to be classpath adjustments to ensure that all the right libraries are loaded.
</blockquote></details>

We need to this for Kantar is its a high priority item for them. If lieu of actually figuring out the core issue, can we either:
* Merge as-is with a todo to update it later
* Make a copy of the s3storageclient that is only used by EMR for now.... like a EMRS3StorageClient that lives in cross-media-measurement?

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/252)* status: all files reviewed, 1 unresolved discussion (waiting on @sandip80 and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/aws/s3/S3StorageClient.kt` line 164 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/252#-NzdbouFDRxjLQtmrKzJ:-O0QvNqwD6mQT12p2-Cu:b7tns87) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/c30c552f78ac0fb0a33b8b635c44d51e2ff23ce3/src/main/kotlin/org/wfanet/measurement/aws/s3/S3StorageClient.kt#L164)):*

We need to this for Kantar is its a high priority item for them. If lieu of actually figuring out the core issue, can we either:
* Merge as-is with a todo to update it later
* Make a copy of the s3storageclient that is only used by EMR for now.... like a EMRS3StorageClient that lives in cross-media-measurement?
</blockquote></details>

Neither of these options is viable, IMO. No workaround for the library issue should be merged into the Halo codebase, whether EMR-specific or not. Whether the current manual approach for EMR also has library dependency issues is irrelevant for this point.

I've now downloaded the EMR Spark image and dug through it. There's a large number of dependency JARs on the classpath, many of which can conflict with our own dependencies. I believe the right approach is to ensure that we're always using our own dependencies rather than the ones from the image. This may involve using a custom image, as described in https://docs.aws.amazon.com/emr/latest/EMR-Serverless-UserGuide/application-custom-image.html. It's also possible there's configuration to adjust the classpath to load everything from our own JAR.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/252)* status: all files reviewed, 1 unresolved discussion (waiting on @sandip80 and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/aws/s3/S3StorageClient.kt` line 164 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/252#-NzdbouFDRxjLQtmrKzJ:-O0R-4TN9iJBhaJqK4Y2:b-i9u3gq) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/c30c552f78ac0fb0a33b8b635c44d51e2ff23ce3/src/main/kotlin/org/wfanet/measurement/aws/s3/S3StorageClient.kt#L164)):*

Neither of these options is viable, IMO. No workaround for the library issue should be merged into the Halo codebase, whether EMR-specific or not. Whether the current manual approach for EMR also has library dependency issues is irrelevant for this point.

I've now downloaded the EMR Spark image and dug through it. There's a large number of dependency JARs on the classpath, many of which can conflict with our own dependencies. I believe the right approach is to ensure that we're always using our own dependencies rather than the ones from the image. This may involve using a custom image, as described in https://docs.aws.amazon.com/emr/latest/EMR-Serverless-UserGuide/application-custom-image.html. It's also possible there's configuration to adjust the classpath to load everything from our own JAR.
</blockquote></details>

A short-term fix I can think of would be to shade the AWS SDK.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/252)* status: all files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/aws/s3/S3StorageClient.kt` line 164 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/252#-NzdbouFDRxjLQtmrKzJ:-O1OLJRuEpKEMxnMLygb:b-2b674n) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/c30c552f78ac0fb0a33b8b635c44d51e2ff23ce3/src/main/kotlin/org/wfanet/measurement/aws/s3/S3StorageClient.kt#L164)):*

A short-term fix I can think of would be to shade the AWS SDK.
</blockquote></details>

@sandip80 - were you able to look into shading the aws dependency in common-jvm?

sandip80:
*[`src/main/kotlin/org/wfanet/measurement/aws/s3/S3StorageClient.kt` line 164 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/252#-NzdbouFDRxjLQtmrKzJ:-O5UjVmm62fOOQrsvxrB:b-66lk05) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/c30c552f78ac0fb0a33b8b635c44d51e2ff23ce3/src/main/kotlin/org/wfanet/measurement/aws/s3/S3StorageClient.kt#L164)):*

@sandip80 - were you able to look into shading the aws dependency in common-jvm?
</blockquote></details>

@SanjayVas @stevenwarejones I think we are good review and merge this. Please let me know if there are any concerns here.

SanjayVas (approved):
> Reviewed 2 of 2 files at r3, 2 of 2 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/252)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @stevenwarejones)

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/252)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @sandip80)

PR #254, opened 2024-06-12
Handle retriable exception of gcs.

renjiezh:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/254)

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/254)* status: all files reviewed, 2 unresolved discussions (waiting on @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/gcloud/gcs/GcsStorageClient.kt` line 80 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/254#-O-D4oRd-CFdgVhhyOcA:-O-D4oRd-CFdgVhhyOcB:b-uugsil) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/2f0ef3bfa7f045549fd426d775867597b6af74ee/src/main/kotlin/org/wfanet/measurement/gcloud/gcs/GcsStorageClient.kt#L80)):*
> ```Kotlin
>         val message = "Fail to write blob."
>         if (e.isRetryable) {
>           throw Status.UNAVAILABLE.withDescription(message).withCause(e).asRuntimeException()
> ```

We can just do the retries here rather than bubbling up to calling code.

That said, it's probably still a good idea to wrap any implementation-specific exception types we would bubble up (e.g. non-retryable errors) into types that are specific to the StorageClient interface. This should not be a gRPC exception as this code should not have a public gRPC dependency. The exception types should be documented on the interface.
___
*[`src/main/kotlin/org/wfanet/measurement/gcloud/gcs/GcsStorageClient.kt` line 84 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/254#-O-D5fNu8a8GJ1QYNdgg:-O-D5fNu8a8GJ1QYNdgh:bdlgrpk) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/2f0ef3bfa7f045549fd426d775867597b6af74ee/src/main/kotlin/org/wfanet/measurement/gcloud/gcs/GcsStorageClient.kt#L84)):*
> ```Kotlin
>           throw Status.UNKNOWN.withDescription(message).withCause(e).asRuntimeException()
>         }
>       } catch (e: Exception) {
> ```

Leave the catch block off if the only behavior is re-throwing.

renjiezh (commented):
Dependent on #255.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/254#-:-O-DX60s3utuo0y9MBk7:b-qtdios)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/gcloud/gcs/GcsStorageClient.kt` line 80 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/254#-O-D4oRd-CFdgVhhyOcA:-O-DWZi7BZqFhM4J05al:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/2f0ef3bfa7f045549fd426d775867597b6af74ee/src/main/kotlin/org/wfanet/measurement/gcloud/gcs/GcsStorageClient.kt#L80)):*

We can just do the retries here rather than bubbling up to calling code.

That said, it's probably still a good idea to wrap any implementation-specific exception types we would bubble up (e.g. non-retryable errors) into types that are specific to the StorageClient interface. This should not be a gRPC exception as this code should not have a public gRPC dependency. The exception types should be documented on the interface.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/gcloud/gcs/GcsStorageClient.kt` line 84 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/254#-O-D5fNu8a8GJ1QYNdgg:-O-DWZTa4DU5qCuR0CnE:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/2f0ef3bfa7f045549fd426d775867597b6af74ee/src/main/kotlin/org/wfanet/measurement/gcloud/gcs/GcsStorageClient.kt#L84)):*

Leave the catch block off if the only behavior is re-throwing.
</blockquote></details>

Done.

SanjayVas (changes requested):
> Reviewed 4 of 4 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/254)* status: all files reviewed, 4 unresolved discussions (waiting on @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/gcloud/gcs/GcsStorageClient.kt` line 67 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/254#-O-DZFYz2kSTi47psCzm:-O-DZFYz2kSTi47psCzn:b45h) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/e788356373e6501e5c61e60a07ca367d9faf2cdb/src/main/kotlin/org/wfanet/measurement/gcloud/gcs/GcsStorageClient.kt#L67)):*
> ```Kotlin
>   override suspend fun writeBlob(blobKey: String, content: Flow<ByteString>): StorageClient.Blob {
>     var attempt = 1
>     while (true) {
> ```





_Suggestion:_
```Kotlin
isActive
```
___
*[`src/main/kotlin/org/wfanet/measurement/gcloud/gcs/GcsStorageClient.kt` line 84 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/254#-O-D_DJR3FFQbZIFOJTh:-O-D_DJR3FFQbZIFOJTi:b2ni8l0) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/e788356373e6501e5c61e60a07ca367d9faf2cdb/src/main/kotlin/org/wfanet/measurement/gcloud/gcs/GcsStorageClient.kt#L84)):*
> ```Kotlin
>   }
> 
>   private suspend fun writeBlobHelper(blobKey: String, content: Flow<ByteString>) =
> ```

nit



_Suggestion:_
```Kotlin
writeBlobInternal
```
___
*[`src/main/kotlin/org/wfanet/measurement/gcloud/gcs/GcsStorageClient.kt` line 84 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/254#-O-D_LA14O_WmI7wqYQz:-O-D_LA14O_WmI7wqYR-:b-t832jy) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/e788356373e6501e5c61e60a07ca367d9faf2cdb/src/main/kotlin/org/wfanet/measurement/gcloud/gcs/GcsStorageClient.kt#L84)):*
> ```Kotlin
>   }
> 
>   private suspend fun writeBlobHelper(blobKey: String, content: Flow<ByteString>) =
> ```

nit: document that this throws GCS StorageException.
___
*[`src/main/kotlin/org/wfanet/measurement/gcloud/gcs/GcsStorageClient.kt` line 135 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/254#-O-DZjmVEuNWiHY_txLk:-O-DZjmVEuNWiHY_txLl:bafww2e) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/e788356373e6501e5c61e60a07ca367d9faf2cdb/src/main/kotlin/org/wfanet/measurement/gcloud/gcs/GcsStorageClient.kt#L135)):*
> ```Kotlin
>    * the transient error of the last attempt.
>    */
>   class MaxAttemptsReachedWritingError(message: String? = null, cause: Throwable? = null) :
> ```

1. Define and document these on the StorageClient interface, not on the GcsStorageClient impl.
2. Use "Exception" in the name rather than "Error", as these inherit from `Exception` and not `Error`.
3. Have some base Exception these inherit from, e.g. defining your own `StorageException`.

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/254)* status: 3 of 5 files reviewed, 2 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/gcloud/gcs/GcsStorageClient.kt` line 67 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/254#-O-DZFYz2kSTi47psCzm:-O-DeZYKEQOo7ENwDwX1:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/e788356373e6501e5c61e60a07ca367d9faf2cdb/src/main/kotlin/org/wfanet/measurement/gcloud/gcs/GcsStorageClient.kt#L67)):*
> ```Kotlin
>   override suspend fun writeBlob(blobKey: String, content: Flow<ByteString>): StorageClient.Blob {
>     var attempt = 1
>     while (true) {
> ```

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/gcloud/gcs/GcsStorageClient.kt` line 135 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/254#-O-DZjmVEuNWiHY_txLk:-O-DesI40XIo-Zu0BwNF:bmfxpif) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/e788356373e6501e5c61e60a07ca367d9faf2cdb/src/main/kotlin/org/wfanet/measurement/gcloud/gcs/GcsStorageClient.kt#L135)):*

1. Define and document these on the StorageClient interface, not on the GcsStorageClient impl.
2. Use "Exception" in the name rather than "Error", as these inherit from `Exception` and not `Error`.
3. Have some base Exception these inherit from, e.g. defining your own `StorageException`.
</blockquote></details>

Done.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/254)* status: all files reviewed, 4 unresolved discussions (waiting on @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/gcloud/gcs/GcsStorageClient.kt` line 63 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/254#-O-DpinR37CL_Np8r0jT:-O-DpinS9BaDFx_nVfdf:b-854e4q) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/55c0d7953fdfaadc9ebe2bdf8f3abe0ecf741df5/src/main/kotlin/org/wfanet/measurement/gcloud/gcs/GcsStorageClient.kt#L63)):*
> ```Kotlin
>   /** Write [content] into a blob in storage. */
>   override suspend fun writeBlob(blobKey: String, content: Flow<ByteString>): StorageClient.Blob {
>     withContext(blockingContext + CoroutineName("writeBlob")) {
> ```

The blockingContext should be inside writeBlobInternal, as that's what's calling the blocking code.

Basically any time you're within a coroutine scope (e.g. in a suspend function), blocking code needs to be wrapped in a context that uses a coroutine dispatcher suitable for blocking calls.
___
*[`src/main/kotlin/org/wfanet/measurement/storage/StorageClient.kt` line 69 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/254#-O-DqgaK-PvSaF3lyFZ3:-O-DqgaLE3ECD-fXhAg4:undefined) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/55c0d7953fdfaadc9ebe2bdf8f3abe0ecf741df5/src/main/kotlin/org/wfanet/measurement/storage/StorageClient.kt#L69)):*
> ```Kotlin
> 
> /**
>  * [Exception] which indicates the max attempts has been reached. [message] and [cause] reflect the
> ```





_Suggestion:_
```Kotlin
[StorageException]
```
___
*[`src/main/kotlin/org/wfanet/measurement/storage/StorageClient.kt` line 69 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/254#-O-DsVnF9S7xkoVpjA7-:-O-DsVnF9S7xkoVpjA70:b-6vjygj) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/55c0d7953fdfaadc9ebe2bdf8f3abe0ecf741df5/src/main/kotlin/org/wfanet/measurement/storage/StorageClient.kt#L69)):*
> ```Kotlin
> 
> /**
>  * [Exception] which indicates the max attempts has been reached. [message] and [cause] reflect the
> ```

nit: Just the cause should reflect the error that caused the last retry attempt. The message should be for this exception itself, which could just be null if there's nothing else useful to say.



_Code quote:_
```Kotlin
[message] and [cause] reflect
```
___
*[`src/main/kotlin/org/wfanet/measurement/storage/StorageClient.kt` line 75 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/254#-O-DqmIi1AvKO0PaejiX:-O-DqmIi1AvKO0PaejiY:bsohxut) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/55c0d7953fdfaadc9ebe2bdf8f3abe0ecf741df5/src/main/kotlin/org/wfanet/measurement/storage/StorageClient.kt#L75)):*
> ```Kotlin
>   StorageException(message, cause)
> 
> /** [Exception] which indicates a permanent writing error. */
> ```

nit: the "permanent" vs. "transient" distinction isn't the salient one here. At most, it's non-retryable.

That said, it might be better just to use the base StorageException for this case. That is, just consider it a less specific exception.

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/254)* status: 3 of 5 files reviewed, 2 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/gcloud/gcs/GcsStorageClient.kt` line 63 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/254#-O-DpinR37CL_Np8r0jT:-O-HaOtAEFyfZHKRGwiW:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/55c0d7953fdfaadc9ebe2bdf8f3abe0ecf741df5/src/main/kotlin/org/wfanet/measurement/gcloud/gcs/GcsStorageClient.kt#L63)):*

The blockingContext should be inside writeBlobInternal, as that's what's calling the blocking code.

Basically any time you're within a coroutine scope (e.g. in a suspend function), blocking code needs to be wrapped in a context that uses a coroutine dispatcher suitable for blocking calls.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/storage/StorageClient.kt` line 69 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/254#-O-DqgaK-PvSaF3lyFZ3:-O-HaLnEAN6vMOrPIsXh:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/55c0d7953fdfaadc9ebe2bdf8f3abe0ecf741df5/src/main/kotlin/org/wfanet/measurement/storage/StorageClient.kt#L69)):*
> ```Kotlin
> 
> /**
>  * [Exception] which indicates the max attempts has been reached. [message] and [cause] reflect the
> ```

Done.

SanjayVas (approved):
> Reviewed 2 of 2 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/254)* status: all files reviewed, 1 unresolved discussion (waiting on @renjiezh)

___
*[`src/main/kotlin/org/wfanet/measurement/gcloud/gcs/GcsStorageClient.kt` line 64 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/254#-O-HgeLB5toa6SRy64lS:-O-HgeLB5toa6SRy64lT:bhsffev) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/d304bd60451a9530bf94efa80255c246230693ba/src/main/kotlin/org/wfanet/measurement/gcloud/gcs/GcsStorageClient.kt#L64)):*
> ```Kotlin
>   override suspend fun writeBlob(blobKey: String, content: Flow<ByteString>): StorageClient.Blob {
>     var attempt = 1
>     while (true) {
> ```

Ahh, I see what happened. Suspend functions don't give access to their CoroutineScope, so you can't call isActive. The equivalent is to just stick a `yield()` call at the start of the loop, but you're already immediately calling a suspend function so it's not needed.
___
*[`src/main/kotlin/org/wfanet/measurement/gcloud/gcs/GcsStorageClient.kt` line 69 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/254#-O-HhR-E3mxLQgNpPfy0:-O-HhR-F5yon5mAMXdUG:b-7kekwo) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/d304bd60451a9530bf94efa80255c246230693ba/src/main/kotlin/org/wfanet/measurement/gcloud/gcs/GcsStorageClient.kt#L69)):*
> ```Kotlin
>       } catch (e: GcsStorageException) {
>         if (!e.isRetryable) {
>           throw StorageException("Fail to write blob due to non-retriable error", e)
> ```

nit: both terms are considered valid, but we use "retryable" more than "retriable".



_Suggestion:_
```Kotlin
non-retryable
```

stevenwarejones (approved):
> Reviewed 3 of 4 files at r4, 1 of 2 files at r6, 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/254)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @renjiezh)

PR #266, opened 2024-08-20
refactor: Modernize usage of Tink

tholenst:
KeysetReader/Writer objects will go away. We use TinkProtoKeysetFormat instead.

This is a pure refactoring.

### DIFF
--- a/src/main/kotlin/org/wfanet/measurement/common/crypto/tink/KeyHandle.kt
+++ b/src/main/kotlin/org/wfanet/measurement/common/crypto/tink/KeyHandle.kt
@@ -33,7 +33,7 @@ class TinkPublicKeyHandle internal constructor(internal val keysetHandle: Keyset
   constructor(serializedKeyset: ByteString) : this(parseKeyset(serializedKeyset))
 
   fun toByteString(): ByteString {
-    return ByteString.copyFrom(TinkProtoKeysetFormat.serializeKeysetWithoutSecret(keysetHandle));
+    return ByteString.copyFrom(TinkProtoKeysetFormat.serializeKeysetWithoutSecret(keysetHandle))
   }
 
   override fun hybridEncrypt(plaintext: ByteString, contextInfo: ByteString?): ByteString {
@@ -48,7 +48,7 @@ class TinkPublicKeyHandle internal constructor(internal val keysetHandle: Keyset
     }
 
     private fun parseKeyset(serialized: ByteString): KeysetHandle {
-      return TinkProtoKeysetFormat.parseKeysetWithoutSecret(serialized.toByteArray());
+      return TinkProtoKeysetFormat.parseKeysetWithoutSecret(serialized.toByteArray())
     }
   }
 }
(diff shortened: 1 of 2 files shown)


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/266)

SanjayVas (approved):
> KeysetReader/Writer objects will go away.

Can you link to any public documentation on this? I don't see any deprecation warnings on the interfaces.

> Reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/266#-:-O5PcUh11h_CAjaRrfiN:b461exf)* status: all files reviewed, 1 unresolved discussion (waiting on @tholenst)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/266#-O5PgErh0SWdGIwr-rpf:-O5PgErh0SWdGIwr-rpg:bhhmmx5) (no related file):*
Approved pending lint fixes.

Note that my approval is not sufficient to merge this PR. This is because we have a two-organization change policy, where every code change must involve at least two member organizations. Since we're both Googlers, there's no other organization yet involved. I can add an external reviewer for you and shepherd this through to the downstream repos once the linter errors are fixed.

___
___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/tink/KeyHandle.kt` line 36 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/266#-O5Pg9cI3GXFDxxHSnJK:-O5Pg9cI3GXFDxxHSnJL:b71orr2) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/ca9c2d77a429849fac17f58f559d06aabd9359fe/src/main/kotlin/org/wfanet/measurement/common/crypto/tink/KeyHandle.kt#L36)):*
> ```Kotlin
> 
>   fun toByteString(): ByteString {
>     return ByteString.copyFrom(TinkProtoKeysetFormat.serializeKeysetWithoutSecret(keysetHandle));
> ```

It's unfortunate that the new interface doesn't offer a way to avoid the extra copies :(

tholenst:
> > KeysetReader/Writer objects will go away.
> 
> Can you link to any public documentation on this? I don't see any deprecation warnings on the interfaces.

The Tink team currently has the policy that all Google internal usages need to be removed before we add deprecation warnings. This is to avoid spamming deprecation warnings on things which we never remove anyhow.

This is the last instance of this API within Google.

> It's unfortunate that the new interface doesn't offer a way to avoid the extra copies :(

In a sense yes. The problem is that we don't want to require a Protobuf dependency on the main part of Tink, so we opted for `byte[]`. But yes, it has disadvantages too :(

tholenst:
@SanjayVas: would you mind adopting this PR instead? I think if one has everything set up (with the linting) and if one is familiar with git things become much more efficient. Since I don't have the linting set up locally and it seems I cannot run the check myself and since I'm not too familiar with git things become a bit more difficult.

WDYT?

SanjayVas:
@tholenst I was going to suggest that as it seemed like you were having some difficulty.

SanjayVas:
Created #269 to replace this.

PR #291, opened 2024-12-18
fix: Postgres connection pool

tristanvuong2021:
Update connection factory for test rule and modify usage of close to deal with potential exceptions.


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/291)

PR #296, opened 2025-01-07
build StorageClient based on blob url

stevenwarejones:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296)

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @Marco-Premier and @stevenwarejones)

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296#-OG1UBjW8gMmOxzY4yBJ:-OG1UBjW8gMmOxzY4yBK:bz6m6c1) (no related file):*
Historically we've avoided having a single binary that depends on multiple cloud provider SDKs. Instead, we've had separate binaries for each provider. This has worked because the operator owns the storage.

Do we have a known use case for supporting multiple storage backends at runtime? e.g. buckets that are not owned by the operator.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @Marco-Premier and @SanjayVas)

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296#-OG1UBjW8gMmOxzY4yBJ:-OG1nZIC4z13IdYNPzet:b-dde71j) (no related file):*

Historically we've avoided having a single binary that depends on multiple cloud provider SDKs. Instead, we've had separate binaries for each provider. This has worked because the operator owns the storage.

Do we have a known use case for supporting multiple storage backends at runtime? e.g. buckets that are not owned by the operator.
</blockquote></details>

As discussed, I removed s3 support for now, but we can easily add it back in later.

SanjayVas (changes requested):
> Reviewed 1 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296)* status: 1 of 4 files reviewed, 4 unresolved discussions (waiting on @kungfucraig, @Marco-Premier, and @stevenwarejones)

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296#-OG1UBjW8gMmOxzY4yBJ:-OG1odNF4X2zfBk5tF9d:b-jowb0y) (no related file):*

As discussed, I removed s3 support for now, but we can easily add it back in later.
</blockquote></details>

For completeness, there is indeed an expected use case of having buckets that are not owned by the operator. In particular, EDPs may own their own buckets for use with the EDP Aggregator.
___
*[`src/main/kotlin/org/wfanet/measurement/storage/Selector.kt` line 32 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296#-OG1oqFj3t83okzDORyD:-OG1oqFj3t83okzDORyE:b53il04) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/f762f016b585dac6a19d5f636ba96dd04af865ce/src/main/kotlin/org/wfanet/measurement/storage/Selector.kt#L32)):*
> ```Kotlin
> )
> 
> fun parseBlobUrl(url: String): BlobUrl? {
> ```

Don't define top-level non-extension functions. Wrap this in a class or object instead. We have some legacy code that does this, but new code should not.

See https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/docs/code-style.md#kotlin

Maybe `StorageClientFactory` with a single public method? That would be clearer than "selector", as a factory is a known pattern.
___
*[`src/main/kotlin/org/wfanet/measurement/storage/Selector.kt` line 34 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296#-OG1pOM13XBLIMBd7YQo:-OG1pOM13XBLIMBd7YQp:bsg7wx5) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/f762f016b585dac6a19d5f636ba96dd04af865ce/src/main/kotlin/org/wfanet/measurement/storage/Selector.kt#L34)):*
> ```Kotlin
> fun parseBlobUrl(url: String): BlobUrl? {
>   // Define regex for different blob URL patterns
>   val s3Regex =
> ```

Use kotlin.text.Regex instead of java.util.regex.Pattern. Also define these as constants (you won't be able to use the `const` keyword, but it's still a constant).
___
*[`src/main/kotlin/org/wfanet/measurement/storage/Selector.kt` line 39 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296#-OG1qP0v3UK8JNFFCHq1:-OG1qP0v3UK8JNFFCHq2:b-k494sk) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/f762f016b585dac6a19d5f636ba96dd04af865ce/src/main/kotlin/org/wfanet/measurement/storage/Selector.kt#L39)):*
> ```Kotlin
>     Pattern.compile("gs://(?<bucket>[^/]+)/(?<path>[^?]+)(?:\\?project=(?<project>[^&]+))?")
>   val fileRegex = Pattern.compile("file://(?<path>.+)")
>   // Match input URL with the regex
> ```

Rather than using regular expressions up-front, I think it'll be clearer to parse into a java.net.URI instance. Then you can just check the `scheme` to know which StorageClient impl to use. After that, you can further parse out the information you need from the URI components.
___
*[`src/main/kotlin/org/wfanet/measurement/storage/Selector.kt` line 78 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296#-OG1rQ349Uctx6isvcn7:-OG1rQ35-jw9ioOfyZCe:bl3q1n3) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/f762f016b585dac6a19d5f636ba96dd04af865ce/src/main/kotlin/org/wfanet/measurement/storage/Selector.kt#L78)):*
> ```Kotlin
>       }
>       "file" -> {
>         FileSystemStorageClient(directory = File("/"))
> ```

I don't think we ever want to allow writing/reading from the true root of the filesystem. Take in a `fileSystemRoot: File` in the constructor instead.



_Code quote:_
```Kotlin
File("/")
```

kungfucraig (dismissed):
> Reviewed 1 of 4 files at r1, 1 of 3 files at r3.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296)* status: 2 of 4 files reviewed, 6 unresolved discussions (waiting on @Marco-Premier, @SanjayVas, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/storage/Selector.kt` line 32 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296#-OG1oqFj3t83okzDORyD:-OGHHajx4we4AcLNRmeX:b-5mvx7m) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/f762f016b585dac6a19d5f636ba96dd04af865ce/src/main/kotlin/org/wfanet/measurement/storage/Selector.kt#L32)):*

Don't define top-level non-extension functions. Wrap this in a class or object instead. We have some legacy code that does this, but new code should not.

See https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/docs/code-style.md#kotlin

Maybe `StorageClientFactory` with a single public method? That would be clearer than "selector", as a factory is a known pattern.
</blockquote></details>

Also add some java doc that describes what is supported, etc.
___
*[`src/main/kotlin/org/wfanet/measurement/storage/Selector.kt` line 34 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296#-OG1pOM13XBLIMBd7YQo:-OGHGsps941Sq1kYkhNa:bt5gl2w) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/f762f016b585dac6a19d5f636ba96dd04af865ce/src/main/kotlin/org/wfanet/measurement/storage/Selector.kt#L34)):*

Use kotlin.text.Regex instead of java.util.regex.Pattern. Also define these as constants (you won't be able to use the `const` keyword, but it's still a constant).
</blockquote></details>

Even though this function won't be called frequently I'd still consider moving them outside of this function so that you only have to pay to compile the regex once. (If indeed you stick with regexes.)
___
*[`src/main/kotlin/org/wfanet/measurement/storage/Selector.kt` line 64 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296#-OGHHQ_M085FD_EiQOkp:-OGHHQ_M085FD_EiQOkq:bu351hx) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/f762f016b585dac6a19d5f636ba96dd04af865ce/src/main/kotlin/org/wfanet/measurement/storage/Selector.kt#L64)):*
> ```Kotlin
>     )
>   }
>   return null
> ```

Throw an exception if the path doesn't match anything that is supported, that is unless you put this in a class/object in which case you can make it private so that only the getStorageClient function can use it. Then the null would be fine.
___
*[`src/main/kotlin/org/wfanet/measurement/storage/BUILD.bazel` line 40 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296#-OGHGG2W7IF_9PLMoFft:-OGHGG2W7IF_9PLMoFfu:bp8pb49) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/f762f016b585dac6a19d5f636ba96dd04af865ce/src/main/kotlin/org/wfanet/measurement/storage/BUILD.bazel#L40)):*
> ```Starlark
> kt_jvm_library(
>     name = "selector",
>     srcs = ["Selector.kt"],
> ```

Being a little more explicit about the name would be nice. Something like: StorageClientSelector.kt

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296)* status: 2 of 4 files reviewed, 6 unresolved discussions (waiting on @kungfucraig, @Marco-Premier, and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/storage/BUILD.bazel` line 40 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296#-OGHGG2W7IF_9PLMoFft:-OL0XvBn4wUqumu5ttwj:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/f762f016b585dac6a19d5f636ba96dd04af865ce/src/main/kotlin/org/wfanet/measurement/storage/BUILD.bazel#L40)):*
<details><summary><i>Previously, kungfucraig (Craig Wright) wrote…</i></summary><blockquote>

Being a little more explicit about the name would be nice. Something like: StorageClientSelector.kt
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/storage/Selector.kt` line 32 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296#-OG1oqFj3t83okzDORyD:-OL0YZrSCNIo_UiJF7bP:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/f762f016b585dac6a19d5f636ba96dd04af865ce/src/main/kotlin/org/wfanet/measurement/storage/Selector.kt#L32)):*
<details><summary><i>Previously, kungfucraig (Craig Wright) wrote…</i></summary><blockquote>

Also add some java doc that describes what is supported, etc.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/storage/Selector.kt` line 34 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296#-OG1pOM13XBLIMBd7YQo:-OL0YiYACdVWt8Sruwm3:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/f762f016b585dac6a19d5f636ba96dd04af865ce/src/main/kotlin/org/wfanet/measurement/storage/Selector.kt#L34)):*
<details><summary><i>Previously, kungfucraig (Craig Wright) wrote…</i></summary><blockquote>

Even though this function won't be called frequently I'd still consider moving them outside of this function so that you only have to pay to compile the regex once. (If indeed you stick with regexes.)
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/storage/Selector.kt` line 39 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296#-OG1qP0v3UK8JNFFCHq1:-OL0YjdP8eW7GXTPhgJI:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/f762f016b585dac6a19d5f636ba96dd04af865ce/src/main/kotlin/org/wfanet/measurement/storage/Selector.kt#L39)):*

Rather than using regular expressions up-front, I think it'll be clearer to parse into a java.net.URI instance. Then you can just check the `scheme` to know which StorageClient impl to use. After that, you can further parse out the information you need from the URI components.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/storage/Selector.kt` line 64 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296#-OGHHQ_M085FD_EiQOkp:-OL0Z28-25n8ifpeKLor:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/f762f016b585dac6a19d5f636ba96dd04af865ce/src/main/kotlin/org/wfanet/measurement/storage/Selector.kt#L64)):*
<details><summary><i>Previously, kungfucraig (Craig Wright) wrote…</i></summary><blockquote>

Throw an exception if the path doesn't match anything that is supported, that is unless you put this in a class/object in which case you can make it private so that only the getStorageClient function can use it. Then the null would be fine.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/storage/Selector.kt` line 78 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296#-OG1rQ349Uctx6isvcn7:-OL0Z3-C3dV5iaT01jsg:bs7wumx) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/f762f016b585dac6a19d5f636ba96dd04af865ce/src/main/kotlin/org/wfanet/measurement/storage/Selector.kt#L78)):*

I don't think we ever want to allow writing/reading from the true root of the filesystem. Take in a `fileSystemRoot: File` in the constructor instead.
</blockquote></details>

Done.

SanjayVas (changes requested):
> Reviewed 2 of 5 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296)* status: 2 of 5 files reviewed, 9 unresolved discussions (waiting on @kungfucraig and @Marco-Premier)

___
*[`src/main/kotlin/org/wfanet/measurement/storage/StorageClientFactory.kt` line 24 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296#-OL1ADjk6Epweb5LAGib:-OL1ADjl-DMt0Yq0tFxc:b45h) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/e4b692efd8825e1c80b34bbe230894ad961744b4/src/main/kotlin/org/wfanet/measurement/storage/StorageClientFactory.kt#L24)):*
> ```Kotlin
> 
> // Data class to store parsed information
> data class BlobUrl(
> ```





_Suggestion:_
```Kotlin
BlobUri
```
___
*[`src/main/kotlin/org/wfanet/measurement/storage/StorageClientFactory.kt` line 25 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296#-OL1AqrCD1yFDRhftosA:-OL1AqrCD1yFDRhftosB:b3x9p9r) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/e4b692efd8825e1c80b34bbe230894ad961744b4/src/main/kotlin/org/wfanet/measurement/storage/StorageClientFactory.kt#L25)):*
> ```Kotlin
> // Data class to store parsed information
> data class BlobUrl(
>   val protocol: String,
> ```

nit: Use standard URI terminology



_Suggestion:_
```Kotlin
scheme
```
___
*[`src/main/kotlin/org/wfanet/measurement/storage/StorageClientFactory.kt` line 72 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296#-OL1CgzX55GdxWvIcHIA:-OL1CgzY3QikwIgXzxlh:b-d5h341) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/e4b692efd8825e1c80b34bbe230894ad961744b4/src/main/kotlin/org/wfanet/measurement/storage/StorageClientFactory.kt#L72)):*
> ```Kotlin
>           val bucket = uri.host
>           val path = uri.path.removePrefix("/")
>           val project = uri.query?.split("=")?.let { parts ->
> ```

I'm not sure where you're getting this from. All the documentation I can find indicates that GCS object URIs don't include the project. The format is just `gs://<bucket_name>/<path>`. Bucket names are globally unique.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296)* status: 2 of 5 files reviewed, 9 unresolved discussions (waiting on @kungfucraig, @Marco-Premier, and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/storage/StorageClientFactory.kt` line 72 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296#-OL1CgzX55GdxWvIcHIA:-OLFpviq4e9xTOIs_nss:b2ue88n) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/e4b692efd8825e1c80b34bbe230894ad961744b4/src/main/kotlin/org/wfanet/measurement/storage/StorageClientFactory.kt#L72)):*

I'm not sure where you're getting this from. All the documentation I can find indicates that GCS object URIs don't include the project. The format is just `gs://<bucket_name>/<path>`. Bucket names are globally unique.
</blockquote></details>

I did see that pattern used. I thought we needed the project in order to construct the Gcs Storage  - https://github.com/world-federation-of-advertisers/common-jvm/blob/4c4802d467f0bed616ce120f16631abb087cd6ce/src/main/kotlin/org/wfanet/measurement/gcloud/gcs/GcsFromFlags.kt#L25 or is that not required anymore?

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296)* status: 2 of 5 files reviewed, 9 unresolved discussions (waiting on @kungfucraig, @Marco-Premier, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/storage/StorageClientFactory.kt` line 72 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296#-OL1CgzX55GdxWvIcHIA:-OLG8ZJX6GxZZJLr13Fy:b-fdm1d4) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/e4b692efd8825e1c80b34bbe230894ad961744b4/src/main/kotlin/org/wfanet/measurement/storage/StorageClientFactory.kt#L72)):*

I did see that pattern used. I thought we needed the project in order to construct the Gcs Storage  - https://github.com/world-federation-of-advertisers/common-jvm/blob/4c4802d467f0bed616ce120f16631abb087cd6ce/src/main/kotlin/org/wfanet/measurement/gcloud/gcs/GcsFromFlags.kt#L25 or is that not required anymore?
</blockquote></details>

My understanding is that the project need not be the one that owns the bucket, just one that has access to the bucket. Therefore it can be the project ID of the operator. Regardless, the bucket is not passed in standard GCS object URIs so if you do need it for some reason it will need to be passed separately.

See https://stackoverflow.com/questions/76769180/is-the-project-id-mandatory-when-using-the-gcp-cloud-storage-api

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296)* status: 1 of 5 files reviewed, 8 unresolved discussions (waiting on @kungfucraig, @Marco-Premier, and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/storage/StorageClientFactory.kt` line 72 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296#-OL1CgzX55GdxWvIcHIA:-OLKc_Wa3AKfhfHenNBD:b-v687gc) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/e4b692efd8825e1c80b34bbe230894ad961744b4/src/main/kotlin/org/wfanet/measurement/storage/StorageClientFactory.kt#L72)):*

My understanding is that the project need not be the one that owns the bucket, just one that has access to the bucket. Therefore it can be the project ID of the operator. Regardless, the bucket is not passed in standard GCS object URIs so if you do need it for some reason it will need to be passed separately.

See https://stackoverflow.com/questions/76769180/is-the-project-id-mandatory-when-using-the-gcp-cloud-storage-api
</blockquote></details>

sounds good. didn't know that. added it as a param on build in case someone needs to include it.
___
*[`src/main/kotlin/org/wfanet/measurement/storage/StorageClientFactory.kt` line 24 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296#-OL1ADjk6Epweb5LAGib:-OLKbt0eBvHUdRD3RInx:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/e4b692efd8825e1c80b34bbe230894ad961744b4/src/main/kotlin/org/wfanet/measurement/storage/StorageClientFactory.kt#L24)):*
> ```Kotlin
> 
> // Data class to store parsed information
> data class BlobUrl(
> ```

Done.

SanjayVas (changes requested):
> Reviewed 2 of 5 files at r4, 2 of 2 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296)* status: all files reviewed, 14 unresolved discussions (waiting on @kungfucraig and @Marco-Premier)

___
*[`src/test/kotlin/org/wfanet/measurement/storage/StorageClientFactoryTest.kt` line 31 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296#-OLL6x2G37_y6_Tx8cgw:-OLL6x2G37_y6_Tx8cgx:b8ucska) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/9bde939e74f18170c111feacb16dc03d166f392e/src/test/kotlin/org/wfanet/measurement/storage/StorageClientFactoryTest.kt#L31)):*
> ```Kotlin
> class StorageClientFactoryTest {
>   @Test
>   fun testS3Url() {
> ```

Use descriptive names. The general pattern is "SUT does X [when Y]"

 taking advantage of the fact that Kotlin allows more flexibility using backticks. For Java, you'd apply the same concept with underscores, e.g. `parseBlobUri_throwsIllegalArgumentException_whenSchemeIsS3`



_Suggestion:_
```Kotlin
`parseBlobUri throws IllegalArgumentException when scheme is s3`
```
___
*[`src/test/kotlin/org/wfanet/measurement/storage/StorageClientFactoryTest.kt` line 36 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296#-OLL7_lO8I46XJWEPJW2:-OLL7_lP1pnbPzfTvnC6:b-tfg173) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/9bde939e74f18170c111feacb16dc03d166f392e/src/test/kotlin/org/wfanet/measurement/storage/StorageClientFactoryTest.kt#L36)):*
> ```Kotlin
>     val blobUrl = BlobUri("s3", "my-bucket", "us-west-2", "path/to/file")
>     assertThrows(IllegalArgumentException::class.java) {
>       StorageClientFactory(blobUrl, Files.createTempDirectory(null).toFile()).build()
> ```

Separate test methods for different SUT methods. Consider whether parseBlobUri should even be part of the public interface, or whether you just want to test the StorageClientFactory build method
___
*[`src/main/kotlin/org/wfanet/measurement/storage/StorageClientFactory.kt` line 24 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296#-OLL0WOJ-z9uR6Q9RrIG:-OLL0WOJ-z9uR6Q9RrIH:bxrspd2) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/9bde939e74f18170c111feacb16dc03d166f392e/src/main/kotlin/org/wfanet/measurement/storage/StorageClientFactory.kt#L24)):*
> ```Kotlin
> 
> // Data class to store parsed information
> data class BlobUri(val scheme: String, val bucket: String?, val region: String?, val path: String)
> ```

It looks like this is never used and always null. Can we drop it for now and add it back in as an optional field if it's needed? I assume it's for S3 specifically, so we can add it if/when S3 support is added.



_Code quote:_
```Kotlin
val region: String?
```
___
*[`src/main/kotlin/org/wfanet/measurement/storage/StorageClientFactory.kt` line 24 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296#-OLL0dHu5VhoF18cpBaO:-OLL0dHu5VhoF18cpBaP:b-4zwbyi) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/9bde939e74f18170c111feacb16dc03d166f392e/src/main/kotlin/org/wfanet/measurement/storage/StorageClientFactory.kt#L24)):*
> ```Kotlin
> 
> // Data class to store parsed information
> data class BlobUri(val scheme: String, val bucket: String?, val region: String?, val path: String)
> ```

I think we should just always make this required and treat file URIs as `file:///bucket/path` for this factory. The storage client constructor invocation could then be changed to

```kotlin
FileSystemStorageClient(rootDirectory.toPath().resolve(bucket).toFile())
```



_Code quote:_
```Kotlin
val bucket: String?
```
___
*[`src/main/kotlin/org/wfanet/measurement/storage/StorageClientFactory.kt` line 27 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296#-OLL-FKz4vuAwYlw7fcF:-OLL-FKz4vuAwYlw7fcG:b70h4zo) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/9bde939e74f18170c111feacb16dc03d166f392e/src/main/kotlin/org/wfanet/measurement/storage/StorageClientFactory.kt#L27)):*
> ```Kotlin
> 
> /*
>  * Builds a [StorageClient] from a blob uri. Currently only supports Google Cloud Storage.
> ```

nit: shouldn't this read that it supports file system and GCS?



_Code quote:_
```Kotlin
 only supports Google Cloud Storage.
```
___
*[`src/main/kotlin/org/wfanet/measurement/storage/StorageClientFactory.kt` line 32 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296#-OLL-_pABcZz1hjwY6qV:-OLL-_pABcZz1hjwY6qW:bcyp7kz) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/9bde939e74f18170c111feacb16dc03d166f392e/src/main/kotlin/org/wfanet/measurement/storage/StorageClientFactory.kt#L32)):*
> ```Kotlin
> class StorageClientFactory(
>   private val blobUrl: BlobUri,
>   private val file: File? = null,
> ```

Document these params
___
*[`src/main/kotlin/org/wfanet/measurement/storage/StorageClientFactory.kt` line 32 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296#-OLL1c3I1YFXFuEB4AKW:-OLL1c3I1YFXFuEB4AKX:b-t2y4bp) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/9bde939e74f18170c111feacb16dc03d166f392e/src/main/kotlin/org/wfanet/measurement/storage/StorageClientFactory.kt#L32)):*
> ```Kotlin
> class StorageClientFactory(
>   private val blobUrl: BlobUri,
>   private val file: File? = null,
> ```

nit: Name this so it's clearer what it's for, since it's not just an arbitrary file



_Suggestion:_
```Kotlin
rootDirectory
```
___
*[`src/main/kotlin/org/wfanet/measurement/storage/StorageClientFactory.kt` line 51 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296#-OLL-U620eGy4Qsjc0bp:-OLL-U620eGy4Qsjc0bq:blerv2f) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/9bde939e74f18170c111feacb16dc03d166f392e/src/main/kotlin/org/wfanet/measurement/storage/StorageClientFactory.kt#L51)):*
> ```Kotlin
>         }
>         "file" -> {
>           FileSystemStorageClient(directory = file!!)
> ```

Check if `file` is null and throw an IllegalStateException indicating that file system hasn't been configured. Right now this instead throws a NullPointerException which isn't very helpful.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296)* status: all files reviewed, 12 unresolved discussions (waiting on @kungfucraig, @Marco-Premier, and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/storage/StorageClientFactory.kt` line 24 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296#-OLL0WOJ-z9uR6Q9RrIG:-OLLEswAAvWgFz94Dtjw:bbgylpr) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/9bde939e74f18170c111feacb16dc03d166f392e/src/main/kotlin/org/wfanet/measurement/storage/StorageClientFactory.kt#L24)):*

It looks like this is never used and always null. Can we drop it for now and add it back in as an optional field if it's needed? I assume it's for S3 specifically, so we can add it if/when S3 support is added.
</blockquote></details>

that is correct. we can add it in later.
___
*[`src/main/kotlin/org/wfanet/measurement/storage/StorageClientFactory.kt` line 24 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296#-OLL0dHu5VhoF18cpBaO:-OLLJ9rVEZbfIXFguc2g:btueczo) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/9bde939e74f18170c111feacb16dc03d166f392e/src/main/kotlin/org/wfanet/measurement/storage/StorageClientFactory.kt#L24)):*

I think we should just always make this required and treat file URIs as `file:///bucket/path` for this factory. The storage client constructor invocation could then be changed to

```kotlin
FileSystemStorageClient(rootDirectory.toPath().resolve(bucket).toFile())
```
</blockquote></details>

this feels little unnatural to me. that's now how URI.create parses it. Also, it prevents writing to the root.
___
*[`src/main/kotlin/org/wfanet/measurement/storage/StorageClientFactory.kt` line 32 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296#-OLL-_pABcZz1hjwY6qV:-OLLJ4127_y_fSh7iplu:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/9bde939e74f18170c111feacb16dc03d166f392e/src/main/kotlin/org/wfanet/measurement/storage/StorageClientFactory.kt#L32)):*

Document these params
</blockquote></details>

Done.
___
*[`src/test/kotlin/org/wfanet/measurement/storage/StorageClientFactoryTest.kt` line 31 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296#-OLL6x2G37_y6_Tx8cgw:-OLLHSQRAdaJjQz6KceE:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/9bde939e74f18170c111feacb16dc03d166f392e/src/test/kotlin/org/wfanet/measurement/storage/StorageClientFactoryTest.kt#L31)):*

Use descriptive names. The general pattern is "SUT does X [when Y]"

 taking advantage of the fact that Kotlin allows more flexibility using backticks. For Java, you'd apply the same concept with underscores, e.g. `parseBlobUri_throwsIllegalArgumentException_whenSchemeIsS3`
</blockquote></details>

Done.
___
*[`src/test/kotlin/org/wfanet/measurement/storage/StorageClientFactoryTest.kt` line 36 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296#-OLL7_lO8I46XJWEPJW2:-OLLHQ7_AvnaEj8jmD10:byb0yi8) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/9bde939e74f18170c111feacb16dc03d166f392e/src/test/kotlin/org/wfanet/measurement/storage/StorageClientFactoryTest.kt#L36)):*

Separate test methods for different SUT methods. Consider whether parseBlobUri should even be part of the public interface, or whether you just want to test the StorageClientFactory build method
</blockquote></details>

you need to parse it to know what blob key to try to read

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296)* status: 3 of 5 files reviewed, 12 unresolved discussions (waiting on @kungfucraig, @Marco-Premier, and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/storage/StorageClientFactory.kt` line 51 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296#-OLL-U620eGy4Qsjc0bp:-OLLJc5K1KaFuISg5jzE:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/9bde939e74f18170c111feacb16dc03d166f392e/src/main/kotlin/org/wfanet/measurement/storage/StorageClientFactory.kt#L51)):*

Check if `file` is null and throw an IllegalStateException indicating that file system hasn't been configured. Right now this instead throws a NullPointerException which isn't very helpful.
</blockquote></details>

Done.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296)* status: 3 of 5 files reviewed, 11 unresolved discussions (waiting on @kungfucraig, @Marco-Premier, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/storage/StorageClientFactory.kt` line 24 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296#-OLL0dHu5VhoF18cpBaO:-OLaESSgCf--3kwcBjKn:b8onegi) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/9bde939e74f18170c111feacb16dc03d166f392e/src/main/kotlin/org/wfanet/measurement/storage/StorageClientFactory.kt#L24)):*
> that's now how URI.create parses it

That's only because of conflating the *path* component of a standard URI with the "path" (key) of the blob, when they are not necessarily the same (and in fact are never the same for S3 and GCS URIs). It would be clearer if you renamed `path` in `BlobUri` to `key`.

> Also, it prevents writing to the root.

I don't believe there's a requirement that this factory let you read from arbitrary existing file system directories. We're free to specify that this is for URIs of the form file:///bucket/key, i.e. URIs where the *scheme* is `file` and the *path* must begin with `/<bucket>/` and be followed by the blob key.

For GCS URIs, the bucket is part of the *authority* of the URI since it does not have leading slash.
___
*[`src/main/kotlin/org/wfanet/measurement/storage/StorageClientFactory.kt` line 51 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296#-OLL-U620eGy4Qsjc0bp:-OLaI5cf4SDEM5tyh6KT:b-jxprsg) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/9bde939e74f18170c111feacb16dc03d166f392e/src/main/kotlin/org/wfanet/measurement/storage/StorageClientFactory.kt#L51)):*

Done.
</blockquote></details>

requireNotNull throws an IllegalArgumentException, which is incorrect as it's not the argument to this method that is invalid. You want `checkNotNull` which throws an IllegalStateException
___
*[`src/test/kotlin/org/wfanet/measurement/storage/StorageClientFactoryTest.kt` line 36 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296#-OLL7_lO8I46XJWEPJW2:-OLaKVcv8cKznaVtJc2x:b-6b5kcl) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/9bde939e74f18170c111feacb16dc03d166f392e/src/test/kotlin/org/wfanet/measurement/storage/StorageClientFactoryTest.kt#L36)):*

you need to parse it to know what blob key to try to read
</blockquote></details>

So it sounds like parseBlobUri should be public. If that's the case, then have a separate test method for getting the illegalargumentexception on parseBlobUri and another for the exception thrown by StorageClientFactory (whether it be the constructor or `build` method).
___
*[`src/main/kotlin/org/wfanet/measurement/storage/StorageClientFactory.kt` line 29 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296#-OLaI-0B9_7CS8-S55AR:-OLaI-0B9_7CS8-S55AS:brdkcf6) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/c572d0599cc6f6501b77ac4b9c9c22f246c81b81/src/main/kotlin/org/wfanet/measurement/storage/StorageClientFactory.kt#L29)):*
> ```Kotlin
>  * Builds a [StorageClient] from a blob uri. Currently only supports Google Cloud Storage and File System.
>  * If file system storage should be supported, input a root File were data should be written/read.
>  * @blobUri - the [BlobUri] from which a StorageClient can be built
> ```

Use correct KDoc syntax



_Suggestion:_
```Kotlin
@param blobUri
```

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296)* status: 1 of 6 files reviewed, 12 unresolved discussions (waiting on @kungfucraig, @Marco-Premier, and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/storage/StorageClientFactory.kt` line 24 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296#-OLL0dHu5VhoF18cpBaO:-OLdIjR7290d3QNYrm3b:b-pbh8xr) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/9bde939e74f18170c111feacb16dc03d166f392e/src/main/kotlin/org/wfanet/measurement/storage/StorageClientFactory.kt#L24)):*

> that's now how URI.create parses it

That's only because of conflating the *path* component of a standard URI with the "path" (key) of the blob, when they are not necessarily the same (and in fact are never the same for S3 and GCS URIs). It would be clearer if you renamed `path` in `BlobUri` to `key`.

> Also, it prevents writing to the root.

I don't believe there's a requirement that this factory let you read from arbitrary existing file system directories. We're free to specify that this is for URIs of the form file:///bucket/key, i.e. URIs where the *scheme* is `file` and the *path* must begin with `/<bucket>/` and be followed by the blob key.

For GCS URIs, the bucket is part of the *authority* of the URI since it does not have leading slash.
</blockquote></details>

Done. I don't feel too strongly about this either way.
___
*[`src/main/kotlin/org/wfanet/measurement/storage/StorageClientFactory.kt` line 51 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296#-OLL-U620eGy4Qsjc0bp:-OLdG9389-3R3Z1tY8TP:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/9bde939e74f18170c111feacb16dc03d166f392e/src/main/kotlin/org/wfanet/measurement/storage/StorageClientFactory.kt#L51)):*

requireNotNull throws an IllegalArgumentException, which is incorrect as it's not the argument to this method that is invalid. You want `checkNotNull` which throws an IllegalStateException
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/storage/StorageClientFactory.kt` line 29 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296#-OLaI-0B9_7CS8-S55AR:-OLdFD_I7Ws5Nxx5cCKd:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/c572d0599cc6f6501b77ac4b9c9c22f246c81b81/src/main/kotlin/org/wfanet/measurement/storage/StorageClientFactory.kt#L29)):*

Use correct KDoc syntax
</blockquote></details>

Done.
___
*[`src/test/kotlin/org/wfanet/measurement/storage/StorageClientFactoryTest.kt` line 36 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296#-OLL7_lO8I46XJWEPJW2:-OLejF70E5L7d9ZCafR0:b-9gehes) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/9bde939e74f18170c111feacb16dc03d166f392e/src/test/kotlin/org/wfanet/measurement/storage/StorageClientFactoryTest.kt#L36)):*

So it sounds like parseBlobUri should be public. If that's the case, then have a separate test method for getting the illegalargumentexception on parseBlobUri and another for the exception thrown by StorageClientFactory (whether it be the constructor or `build` method).
</blockquote></details>

refactored this to be a StorageClient and not a factory. I think that simplifies usage which means that parseBlobUri can now be private.
___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296#-OLeiuy6-bjBJNu1L71s:-OLeiuy6-bjBJNu1L71t:b-18356v) (no related file):*
I refactored this PR to not be a factory but instead to just automatically build the storage client needed. I think this better fits the use case.

SanjayVas (changes requested):
> Reviewed 3 of 5 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296)* status: 4 of 6 files reviewed, 11 unresolved discussions (waiting on @kungfucraig, @Marco-Premier, and @stevenwarejones)

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296#-OLeiuy6-bjBJNu1L71s:-OLf3CMmBiYj52SNqvdS:b-t0kbcq) (no related file):*

I refactored this PR to not be a factory but instead to just automatically build the storage client needed. I think this better fits the use case.
</blockquote></details>

That's actually a bit of a strange design, because the StorageClient methods take in the blob key which is already part of the BlobUri taking in the constructor.
___
*[`src/test/kotlin/org/wfanet/measurement/storage/SelectedStorageClientTest.kt` line 55 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296#-OLf2-J73PEaxXqy54uq:-OLf2-J73PEaxXqy54ur:b172ecq) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/7735aa9400b446e4fab797621133716a39a92fd9/src/test/kotlin/org/wfanet/measurement/storage/SelectedStorageClientTest.kt#L55)):*
> ```Kotlin
>   @Test
>   fun `able to write blob from filesystem`() = runBlocking {
>     val fileUri = "file://bucket/path/to"
> ```

File URIs don't have an authority component, and therefore have another slash which begins the path component.



_Suggestion:_
```Kotlin
file:///bucket/path/to
```
___
*[`src/main/kotlin/org/wfanet/measurement/storage/SelectedStorageClient.kt` line 72 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296#-OLf3LF24r0TOK5f1A0-:-OLf3LF24r0TOK5f1A00:b-tsml4u) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/7735aa9400b446e4fab797621133716a39a92fd9/src/main/kotlin/org/wfanet/measurement/storage/SelectedStorageClient.kt#L72)):*
> ```Kotlin
> 
>   override suspend fun getBlob(blobKey: String): StorageClient.Blob? {
>     val path = File(blobUri.key, blobKey).path
> ```

This is incorrect. The key that's part of the blob URI is redundant with the `blobKey` parameter. It should instead be something like

```kotlin
require(blobUri.key == blobKey) { "blobKey does not match" }
return underlyingClient.getBlob(blobKey)
```
___
*[`src/main/kotlin/org/wfanet/measurement/storage/SelectedStorageClient.kt` line 80 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296#-OLf1hYk7j0tF5QvBnM7:-OLf1hYk7j0tF5QvBnM8:bqzxtf5) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/7735aa9400b446e4fab797621133716a39a92fd9/src/main/kotlin/org/wfanet/measurement/storage/SelectedStorageClient.kt#L80)):*
> ```Kotlin
>     private fun parseBlobUri(url: String): BlobUri {
>       val uri = URI.create(url)
>       val bucket = uri.host
> ```

The bucket won't be the host for file URIs, as file URIs do not have an authority component. You need to look at the scheme first before deciding which component the bucket comes from.

Again, it's `file:///<bucket>/<key>` for file URIs (note the three slashes after the colon) and `gs://<bucket>/<key>` for GCS object URIs.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296)* status: 4 of 6 files reviewed, 11 unresolved discussions (waiting on @kungfucraig, @Marco-Premier, and @SanjayVas)

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296#-OLeiuy6-bjBJNu1L71s:-OLfA4mw770qKHtcYNg4:b-kuwvyt) (no related file):*

That's actually a bit of a strange design, because the StorageClient methods take in the blob key which is already part of the BlobUri taking in the constructor.
</blockquote></details>

see my other comment about the blobUri passed in the constructor essentially setting up the storageClient to be a prefixed storage client restricted to reading different blobs with that path using relative blob keys
___
*[`src/main/kotlin/org/wfanet/measurement/storage/SelectedStorageClient.kt` line 72 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296#-OLf3LF24r0TOK5f1A0-:-OLf9RhR5VFATxtCM0Bj:b-6jiymv) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/7735aa9400b446e4fab797621133716a39a92fd9/src/main/kotlin/org/wfanet/measurement/storage/SelectedStorageClient.kt#L72)):*

This is incorrect. The key that's part of the blob URI is redundant with the `blobKey` parameter. It should instead be something like

```kotlin
require(blobUri.key == blobKey) { "blobKey does not match" }
return underlyingClient.getBlob(blobKey)
```
</blockquote></details>

I'm treating the blobUri like a prefixedStorageClient. The storage client is initialized with the blobUri and then reads from files within that path.
I'm thinking of two use cases:
1. Receiver is given the full blob uri to the file. In that case, they just want that blobUri.key
2. Receiver is given a blobUri to the parent folder and expected to read a file from that directory. In that case, they'd pass in the file they want to read within that structure (see the tests)

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296)* status: 3 of 6 files reviewed, 11 unresolved discussions (waiting on @kungfucraig, @Marco-Premier, and @SanjayVas)

___
*[`src/test/kotlin/org/wfanet/measurement/storage/SelectedStorageClientTest.kt` line 55 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296#-OLf2-J73PEaxXqy54uq:-OLfBLYMBEDeRLeXfDB4:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/7735aa9400b446e4fab797621133716a39a92fd9/src/test/kotlin/org/wfanet/measurement/storage/SelectedStorageClientTest.kt#L55)):*

File URIs don't have an authority component, and therefore have another slash which begins the path component.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/storage/SelectedStorageClient.kt` line 80 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296#-OLf1hYk7j0tF5QvBnM7:-OLfBMq76fQmhQHcdWHC:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/7735aa9400b446e4fab797621133716a39a92fd9/src/main/kotlin/org/wfanet/measurement/storage/SelectedStorageClient.kt#L80)):*

The bucket won't be the host for file URIs, as file URIs do not have an authority component. You need to look at the scheme first before deciding which component the bucket comes from.

Again, it's `file:///<bucket>/<key>` for file URIs (note the three slashes after the colon) and `gs://<bucket>/<key>` for GCS object URIs.
</blockquote></details>

Done.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296)* status: 3 of 6 files reviewed, 10 unresolved discussions (waiting on @kungfucraig, @Marco-Premier, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/storage/SelectedStorageClient.kt` line 72 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296#-OLf3LF24r0TOK5f1A0-:-OLfGanR89NYY5lvUVb5:bm0i8pj) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/7735aa9400b446e4fab797621133716a39a92fd9/src/main/kotlin/org/wfanet/measurement/storage/SelectedStorageClient.kt#L72)):*

I'm treating the blobUri like a prefixedStorageClient. The storage client is initialized with the blobUri and then reads from files within that path.
I'm thinking of two use cases:
1. Receiver is given the full blob uri to the file. In that case, they just want that blobUri.key
2. Receiver is given a blobUri to the parent folder and expected to read a file from that directory. In that case, they'd pass in the file they want to read within that structure (see the tests)
</blockquote></details>

Blob storage doesn't actually have a concept of folders. They just happen to allow slashes in keys so you can organize things to look like folders.

You should always expect blob URIs to point to the actual blob. Anything else is an invalid URI.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296)* status: 3 of 6 files reviewed, 10 unresolved discussions (waiting on @kungfucraig, @Marco-Premier, and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/storage/SelectedStorageClient.kt` line 72 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296#-OLf3LF24r0TOK5f1A0-:-OLfHBx0DHaeXQZO4Uit:bjwu5cf) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/7735aa9400b446e4fab797621133716a39a92fd9/src/main/kotlin/org/wfanet/measurement/storage/SelectedStorageClient.kt#L72)):*

Blob storage doesn't actually have a concept of folders. They just happen to allow slashes in keys so you can organize things to look like folders.

You should always expect blob URIs to point to the actual blob. Anything else is an invalid URI.
</blockquote></details>

I was using this as a precedent. https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/62660fdfc522ae6daff4329f893a27e5d9e5675d/src/main/kotlin/org/wfanet/panelmatch/common/storage/PrefixedStorageClient.kt#L40

wdyt if I rename BlobUri.key to BlobUri.keyPrefix ?

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296)* status: 3 of 6 files reviewed, 10 unresolved discussions (waiting on @kungfucraig, @Marco-Premier, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/measurement/storage/SelectedStorageClient.kt` line 72 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296#-OLf3LF24r0TOK5f1A0-:-OLfImmNCFvM4r2A1zvS:b-e4y7b8) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/7735aa9400b446e4fab797621133716a39a92fd9/src/main/kotlin/org/wfanet/measurement/storage/SelectedStorageClient.kt#L72)):*

I was using this as a precedent. https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/62660fdfc522ae6daff4329f893a27e5d9e5675d/src/main/kotlin/org/wfanet/panelmatch/common/storage/PrefixedStorageClient.kt#L40

wdyt if I rename BlobUri.key to BlobUri.keyPrefix ?
</blockquote></details>

If it doesn't point to a blob, it's not a valid blob URI. In that case, it would be the string prefix of a blob URI.

I'd just drop use case 2 entirely, and assume you only ever have real blob URIs. It just happens to be that you only need to parse the scheme and the bucket out of the URI in order to determine how to construct a storage client.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296)* status: 3 of 6 files reviewed, 10 unresolved discussions (waiting on @kungfucraig, @Marco-Premier, and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/storage/SelectedStorageClient.kt` line 72 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296#-OLf3LF24r0TOK5f1A0-:-OLiotKT5cWADcKaXNC5:b-9fub0a) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/7735aa9400b446e4fab797621133716a39a92fd9/src/main/kotlin/org/wfanet/measurement/storage/SelectedStorageClient.kt#L72)):*

If it doesn't point to a blob, it's not a valid blob URI. In that case, it would be the string prefix of a blob URI.

I'd just drop use case 2 entirely, and assume you only ever have real blob URIs. It just happens to be that you only need to parse the scheme and the bucket out of the URI in order to determine how to construct a storage client.
</blockquote></details>

okay - that works. we can revisit later if we have a use case.

SanjayVas (changes requested):
> Reviewed 1 of 5 files at r4, 1 of 5 files at r9, 2 of 2 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296)* status: all files reviewed, 4 unresolved discussions (waiting on @kungfucraig and @Marco-Premier)

___
*[`src/main/kotlin/org/wfanet/measurement/storage/SelectedStorageClient.kt` line 71 at r12](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296#-OLk4x2P1_lgAuW_hFvR:-OLk4x2P1_lgAuW_hFvS:bpbhhjm) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/c1834d39aafe00d3be16083e2f8d7cb9ee82c712/src/main/kotlin/org/wfanet/measurement/storage/SelectedStorageClient.kt#L71)):*
> ```Kotlin
> 
>   override suspend fun getBlob(blobKey: String): StorageClient.Blob? {
>     return underlyingClient.getBlob(blobKey)
> ```

Verify that `blobKey` matches `blobUri.key`, otherwise the caller could get unexpected results.

stevenwarejones (commented):
> Dismissed @kungfucraig from 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296)* status: 4 of 6 files reviewed, 2 unresolved discussions (waiting on @kungfucraig, @Marco-Premier, and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/storage/SelectedStorageClient.kt` line 71 at r12](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/296#-OLk4x2P1_lgAuW_hFvR:-OLkRsKs2q2v6D23HzmT:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/c1834d39aafe00d3be16083e2f8d7cb9ee82c712/src/main/kotlin/org/wfanet/measurement/storage/SelectedStorageClient.kt#L71)):*

Verify that `blobKey` matches `blobUri.key`, otherwise the caller could get unexpected results.
</blockquote></details>

Done.

PR #301, opened 2025-02-04
feat!: Add overload of verifyAndDecodeBearerToken using credentials

SanjayVas:
This allows the caller to specifically handle the case where there is no bearer token in the gRPC headers.

BREAKING CHANGE: OpenIdConnectAuthentication is renamed to OAuthTokenAuthentication for accuracy.


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/301)

PR #317, opened 2025-06-10
fix: Ensure that Spanner AsyncResultSet is cancelled on coroutine cancellation

SanjayVas:
Issue: https://github.com/world-federation-of-advertisers/cross-media-measurement/issues/2376


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/317)

PR #321, opened 2025-06-20
feat: Add InProcessServersMethods utility for in-process gRPC servers

georgi:
This utility provides a convenient method to start in-process gRPC servers with common configuration including logging interceptors and graceful shutdown.

### DIFF
--- a/src/main/kotlin/org/wfanet/measurement/common/grpc/InProcessServersMethods.kt
+++ b/src/main/kotlin/org/wfanet/measurement/common/grpc/InProcessServersMethods.kt
@@ -1,0 +1,64 @@
+/*
+ * Copyright 2023 The Cross-Media Measurement Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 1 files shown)


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/321)

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/321)* status: all files reviewed, 1 unresolved discussion

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/321#-OTT6if45_sLRN3aD4A_:-OTT6if45_sLRN3aD4Aa:b6nazaj) (no related file):*
This appears to be copied from https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/src/main/kotlin/org/wfanet/measurement/reporting/deploy/v2/common/InProcessServersMethods.kt, which means the following must happen:

1. The copyright year must be adjusted to match the original file.
2. The PR which adopts this into the main repo must also refactor usages of the file it was copied from to use this instead.

SanjayVas:
> 1) was advised by @kungfucraig to be updated, so please sync on that

I already spoke with @kungfucraig. He asked you to update it because he was unaware the file was being moved from the main repo, and this PR made it sound like it was a new file.

Issue #335, opened 2025-08-12
Set statement_timeout to prevent long-running queries


tristanvuong2021:
Occasionally, queries take much longer than they should. This should help kill the queries if that occurs.

PR #336, opened 2025-08-12
perf: Add statement_timeout to Postgres Client Connections

tristanvuong2021:
Set statement_timeout to 60s for Postgres connections.

Issue: #335

### DIFF
--- a/src/main/kotlin/org/wfanet/measurement/common/db/r2dbc/postgres/PostgresDatabaseClient.kt
+++ b/src/main/kotlin/org/wfanet/measurement/common/db/r2dbc/postgres/PostgresDatabaseClient.kt
@@ -21,6 +21,7 @@ import io.r2dbc.spi.ConnectionFactories
 import io.r2dbc.spi.ConnectionFactory
 import io.r2dbc.spi.ConnectionFactoryOptions
 import io.r2dbc.spi.IsolationLevel
+import java.time.Duration
 import kotlinx.coroutines.reactive.awaitSingle
 import org.wfanet.measurement.common.db.postgres.PostgresFlags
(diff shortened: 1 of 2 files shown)


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/336)

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/336)* status: 0 of 2 files reviewed, 1 unresolved discussion

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/336#-OXVC-KK4TRoSdnay6_B:-OXVC-KK4TRoSdnay6_C:baxyp83) (no related file):*
Does the RPC timeout not inherently take care of this? The client should be written in such a way that the RPC being cancelled results in everything being closed and cleaned up.

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/336)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/336#-OXVC-KK4TRoSdnay6_B:-OXVMuUvErmRGucQOgA8:b-qjhycd) (no related file):*

Does the RPC timeout not inherently take care of this? The client should be written in such a way that the RPC being cancelled results in everything being closed and cleaned up.
</blockquote></details>

It doesn't seem to inherently take care of it since it still occasionally happens. statement_timeout is the only way I can think of to guarantee it since it affects the connection directly on the Postgres side. If the Reporting Server crashes, then any additional handling might not even run.

PR #340, opened 2025-09-04
build: Update netty-handler to 4.1.124.Final

laureanobrs:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/340)

PR #341, opened 2025-09-19
build: Update rules_kotlin_jvm dep for grpc-java

SanjayVas:


SanjayVas:
Depends on world-federation-of-advertisers/rules_kotlin_jvm#16

PR #344, opened 2025-09-24
feat: add overload for envelope encryption with custom Tink keyset parsing

Marco-Premier:
Issue: #343


Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/344)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/tink/WithEnvelopeEncryption.kt` line 36 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/344#-O_xoB91-GuOQ2NH8194:-Oa-efotCNHvwYW-kxvf:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/f75418b6733291adfb4a7e5129a2573c9a5775a1/src/main/kotlin/org/wfanet/measurement/common/crypto/tink/WithEnvelopeEncryption.kt#L36)):*

Since you didn't actually add an overload but instead added an optional parameter, this is no longer always true.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/tink/WithEnvelopeEncryption.kt` line 55 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/344#-O_xpfuu9b8sNTDvPwV8:-Oa-zpM9CS6_nWT6aATj:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/f75418b6733291adfb4a7e5129a2573c9a5775a1/src/main/kotlin/org/wfanet/measurement/common/crypto/tink/WithEnvelopeEncryption.kt#L55)):*

Use the same signature as `TinkProtoKeysetFormat.parseEncryptedKeyset` so you can use that function directly as the default. Otherwise there's no point in having both this function parameter and the three key parameters to `withEnvelopeEncryption`, as the caller could just do the binding themselves.
</blockquote></details>

Done.

SanjayVas (changes requested):
> @SanjayVas reviewed 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/344)* status: all files reviewed, 2 unresolved discussions (waiting on @Marco-Premier)

___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/tink/WithEnvelopeEncryption.kt` line 39 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/344#-Oa14-uw8Ldyu-aeUcMM:-Oa14-uw8Ldyu-aeUcMN:undefined) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/db3af7f6bac8049afc00a46b969a5285c564accd/src/main/kotlin/org/wfanet/measurement/common/crypto/tink/WithEnvelopeEncryption.kt#L39)):*
> ```Kotlin
>  * @param encryptedDek the encrypted data encryption key (DEK). By default this is
>  *        expected to be in Tink binary format, an alternative format (e.g. JSON)
>  *        can be supported by providing a custom [keysetParser].
> ```





_Suggestion:_
```Kotlin
parseEncryptedKeyset
```
___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/tink/WithEnvelopeEncryption.kt` line 64 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/344#-Oa144hJ4WS1XcmmMhck:-Oa144hJ4WS1XcmmMhcl:b2ni8l0) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/db3af7f6bac8049afc00a46b969a5285c564accd/src/main/kotlin/org/wfanet/measurement/common/crypto/tink/WithEnvelopeEncryption.kt#L64)):*
> ```Kotlin
>     encryptedDek.toByteArray(),
>     kekAead,
>     byteArrayOf()
> ```

nit



_Suggestion:_
```Kotlin
null
```

PR #348, opened 2025-10-06
add aead support to WithEnvelopeEncryption

stevenwarejones:


SanjayVas (changes requested):
> @SanjayVas reviewed 4 of 5 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/348)* status: all files reviewed, 1 unresolved discussion (waiting on @Marco-Premier)

___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/tink/AeadStorageClient.kt` line 94 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/348#-OauemA26A8TU1ahqeEO:-OauemA26A8TU1ahqeEP:b-8x0olt) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/bd55ddeabd9414688a5cb491326e3cef6a938878/src/main/kotlin/org/wfanet/measurement/common/crypto/tink/AeadStorageClient.kt#L94)):*
> ```Kotlin
>     override fun read(): Flow<ByteString> {
>       val associatedData: ByteString = blobKey.toByteStringUtf8()
>       val data = runBlocking { blob.read().flatten().toByteArray() }
> ```

`runBlocking` must not be used here. You can suspend inside of the flow collector instead.

e.g.

```kotlin
override fun read(): Flow<ByteString> = flow {
  // Suspend in here
  // Emit what is needed
}
```

In general, we should only see `runBlocking` in tests or the transition from the `main` function into coroutine-aware code.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/348)* status: 4 of 5 files reviewed, 1 unresolved discussion (waiting on @Marco-Premier and @SanjayVas)

___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/tink/AeadStorageClient.kt` line 94 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/348#-OauemA26A8TU1ahqeEO:-Oauocu79yF5vV4Mrsjf:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/bd55ddeabd9414688a5cb491326e3cef6a938878/src/main/kotlin/org/wfanet/measurement/common/crypto/tink/AeadStorageClient.kt#L94)):*

`runBlocking` must not be used here. You can suspend inside of the flow collector instead.

e.g.

```kotlin
override fun read(): Flow<ByteString> = flow {
  // Suspend in here
  // Emit what is needed
}
```

In general, we should only see `runBlocking` in tests or the transition from the `main` function into coroutine-aware code.
</blockquote></details>

Done.

PR #389, opened 2026-06-16
feat: Add writeBlobIfGeneration to ConditionalOperationStorageClient

stevenwarejones:
Adds a generation-keyed conditional blob write primitive (GCS `IfGenerationMatch`) to `ConditionalOperationStorageClient`. Pass `expectedGeneration=0` for first-writer-wins (write-if-absent); pass a non-zero value for compare-and-swap on the current generation. Concurrent writers race atomically at the server: exactly one write lands per `expectedGeneration` value, the others fail-fast with `BlobChangedException`.

Motivated by the memoized VID pipeline in cross-media-measurement: the Phase-0 and Phase-1 TEE apps write multi-GB encrypted blobs to deterministic keys, and Pub/Sub redelivery can spawn concurrent writers that must not clobber each other. The right pattern there is:

```kotlin
// At VM start, before compute:
val priorBlob = storage.getBlob(key)
val expectedGen = priorBlob?.generation ?: 0L

... compute ...

// Blob first, gated on the generation observed at start:
try {
  storage.writeBlobIfGeneration(key, expectedGen, encryptedBytes)
  insertMetadataRow(key, dek, ...)  // AIP-155 request_id; deduped on retry
} catch (e: BlobChangedException) {
  // Another writer won the slot. Don't re-read the generation and retry — that would
  // clobber the winner's bytes and corrupt their row. Just look up the winning row to
  // confirm someone is making progress, then ack.
}
```

That handles every scenario cleanly without per-attempt-UUID workarounds or bucket-lifecycle-policy dependencies:
- **Fresh blob:** `expectedGen=0`, succeeds.
- **Crash after write, before row:** next VM reads gen, overwrites cleanly, inserts row.
- **Concurrent peers:** exactly one wins via the precondition; the loser sees the winner's row and acks.

See world-federation-of-advertisers/cross-media-measurement#3999.

## Side fix: race in the existing `GcsStorageClient.writeBlob`

The old implementation split this into `storage.create()` + `Blob.write(generationMatch(generation))`, which introduced a small window where a concurrent writer between the two ops would cause `writeBlob` to fail with HTTP 412 despite being documented as unconditional. New shape uses `storage.writer(blobInfo)` directly — single round trip, no precondition, truly unconditional.

`writeBlobIfUnchanged` is also restructured to use the writer directly with `generationMatch(blob.generation)` instead of routing through a shared `Blob.write` extension that hardcoded a precondition.

## Wrapper requirements (compile-time)

The encryption and record-IO wrappers (`AeadStorageClient`, `StreamingAeadStorageClient`, `MesosRecordIoStorageClient`) now accept only `ConditionalOperationStorageClient` as their underlying client. Misconfiguration is caught at the wrapper constructor's call site by the Kotlin compiler — no runtime cast, no `IllegalArgumentException` from inside a write call. The corresponding `withEnvelopeEncryption` extension is now on `ConditionalOperationStorageClient` for the same reason.

To keep existing production callers compiling (notably `EncryptedStorage.buildEncryptedMesosStorageClient` through `SelectedStorageClient`, which can return either a `GcsStorageClient` or a `FileSystemStorageClient`), `FileSystemStorageClient` also implements `ConditionalOperationStorageClient` in this PR — generation is the file's `lastModified` timestamp, snapshotted at `Blob` construction. `writeBlob` ensures `lastModified` strictly advances across writes (filesystem millisecond resolution can otherwise alias two writes to the same generation). `writeBlobIfGeneration` uses `Files.createFile` for the absent case (atomic) and a temp+atomic-move for the present case.

## Dedup

While addressing review feedback, several near-duplicate code paths were collapsed:

- `GcsStorageClient`'s three write methods (`writeBlob`, `writeBlobIfUnchanged`, `writeBlobIfGeneration`) shared ~25 lines of nearly-identical `withContext + storage.writer + storage.get + try/catch` skeleton. They now delegate to a single `doWrite(...)` private helper, differing only by the write options, the coroutine name, and the precondition-error message.
- `AeadStorageClient` and `StreamingAeadStorageClient` each had three copies of the `aead.encrypt(content, blobKey)` boilerplate. Now extracted to a private `encrypt(blobKey, content)` helper per class; the three call sites become one-liners.
- `MesosRecordIoStorageClient.writeBlob` was duplicating the size+delimiter framing logic that was already in the `recordIoFraming()` private helper used by the conditional methods. `writeBlob` now reuses the helper.
- `require(expectedGeneration >= 0)` is enforced at the wrapper boundary in addition to the leaf clients, so misuse fails locally regardless of which underlying client is wrapped.

## Surface

- `ConditionalOperationStorageClient.writeBlobIfGeneration(blobKey, expectedGeneration, content)`
- Implementations: `GcsStorageClient`, `FileSystemStorageClient`, `InMemoryStorageClient`, `AeadStorageClient`, `StreamingAeadStorageClient`, `MesosRecordIoStorageClient`

## Tests

- `AbstractConditionalOperationStorageClientTest` is the canonical conformance suite for the conditional-write primitives. It covers both the `expectedGeneration=0` path (write-if-absent: success, throws-when-exists, leaves-blob-untouched on failure, empty content, no-key-lock) AND the CAS-on-existing-blob path (stale-generation throws, matching-generation overwrites, negative-generation rejected). The base class declares a `protected abstract suspend fun getGeneration(blobKey)` so each backend's test exposes generation in the backend-appropriate way; this lets the CAS tests be backend-agnostic.
- Every `ConditionalOperationStorageClient` implementation extends the conformance suite: `GcsStorageClientTest`, `FileSystemStorageClientTest`, `InMemoryStorageClientTest`, `AeadStorageClientTest`, `StreamingAeadStorageClientTest`, `MesosRecordIoStorageClientTest`.
- `listBlobs returns all blobs without duplicates` is also a conformance test, run against every backend.
- Backend-specific tests that genuinely belong per-impl stay put: GCS PATCH-with-null-metadata-deletes-the-key (GCS REST/SDK quirk), `InMemoryStorageClient.contents()` (InMemory-specific accessor), `FileSystemStorageClient` subdirectory write path.

Issue: world-federation-of-advertisers/cross-media-measurement#4019


Marco-Premier (approved):
> @Marco-Premier made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/389)* status: 0 of 11 files reviewed, 3 unresolved discussions (waiting on stevenwarejones).

___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/tink/AeadStorageClient.kt` line 39 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/389#-OvJrWN-5WSwQspUpFvw:-OvJrWN-5WSwQspUpFvx:b8afblb) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/73cd7e1fc81dbe3765cef33789ffc54bf66f0848/src/main/kotlin/org/wfanet/measurement/common/crypto/tink/AeadStorageClient.kt#L39)):*
> ```Kotlin
>  */
> class AeadStorageClient(private val storageClient: StorageClient, private val aead: Aead) :
>   StorageClient, ConditionalOperationStorageClient {
> ```

This wrapper statically declares `ConditionalOperationStorageClient` but only honors it if the wrapped client does, and we only find out via a per-call `require` that throws an undocumented,  `IllegalArgumentException`.

Would it worth to validate once at construction and hold a typed delegate so misconfiguration fails immediately instead of after compute.?

Same apply to `StreamingAeadStorageClient.kt` and `MesosRecordIoStorageClient.kt`

```
class AeadStorageClient(private val storageClient: StorageClient, private val aead: Aead) :

    StorageClient, ConditionalOperationStorageClient {

    private val conditionalDelegate: ConditionalOperationStorageClient =
      requireNotNull(storageClient as? ConditionalOperationStorageClient) {
        "AeadStorageClient requires a ConditionalOperationStorageClient, but wrapped " +
          "${storageClient::class.simpleName} does not support conditional writes"
      }   

      
    // then in the methods, drop the per-call require and use the typed delegate:
    override suspend fun writeBlobIfGeneration(
      blobKey: String,
      expectedGeneration: Long,
      content: Flow<ByteString>,
    ): StorageClient.Blob {

      val ciphertext: ByteString = /* ...encrypt... */

      val wrappedBlob = conditionalDelegate.writeBlobIfGeneration(blobKey, expectedGeneration, flow { emit(ciphertext) })
      return EncryptedBlob(wrappedBlob, blobKey)

    } 
```
___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/tink/AeadStorageClient.kt` line 77 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/389#-OvJs_FxD_a_Ql8BfpqA:-OvJs_FxD_a_Ql8BfpqB:b-xisl04) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/73cd7e1fc81dbe3765cef33789ffc54bf66f0848/src/main/kotlin/org/wfanet/measurement/common/crypto/tink/AeadStorageClient.kt#L77)):*
> ```Kotlin
>     content: Flow<ByteString>,
>   ): StorageClient.Blob {
>     require(storageClient is ConditionalOperationStorageClient) {
> ```

The `expectedGeneration >= 0` guard only exists in the leaf clients, shall we have early failure regardless of which underlying client is wrapped?

e.g.

```
override suspend fun writeBlobIfGeneration(
    blobKey: String,
    expectedGeneration: Long,
    content: Flow<ByteString>,
  ): StorageClient.Blob {

    require(expectedGeneration >= 0) { "expectedGeneration must be >= 0, got $expectedGeneration" }
    // ...existing body...
  }
```

Same apply to `StreamingAeadStorageClient.kt` and `MesosRecordIoStorageClient.kt`
___
*[`src/main/kotlin/org/wfanet/measurement/storage/MesosRecordIoStorageClient.kt` line 55 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/389#-OvJsDT06WcW7d8ZjGyk:-OvJsDT06WcW7d8ZjGyl:b-unhrgy) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/73cd7e1fc81dbe3765cef33789ffc54bf66f0848/src/main/kotlin/org/wfanet/measurement/storage/MesosRecordIoStorageClient.kt#L55)):*
> ```Kotlin
>    * @return A Blob object representing the RecordIO data that was written to storage.
>    */
>   override suspend fun writeBlob(blobKey: String, content: Flow<ByteString>): StorageClient.Blob {
> ```

The framing here is now duplicated by the new `recordIoFraming()` helper (line 117) — I'd suggest to have `writeBlob` reuse it so there's one source of truth for the size+delimiter prefix. 

```
override suspend fun writeBlob(blobKey: String, content: Flow<ByteString>): StorageClient.Blob {

    val wrappedBlob: StorageClient.Blob = storageClient.writeBlob(blobKey, recordIoFraming(content))
    return Blob(wrappedBlob, blobKey)

  } 
```

stevenwarejones (commented):
> @stevenwarejones+AGNT made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/389)* status: 0 of 12 files reviewed, 3 unresolved discussions (waiting on Marco-Premier).

___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/tink/AeadStorageClient.kt` line 39 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/389#-OvJrWN-5WSwQspUpFvw:-OvKznXZ14X_hdGuykBq:b-99u6v5) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/73cd7e1fc81dbe3765cef33789ffc54bf66f0848/src/main/kotlin/org/wfanet/measurement/common/crypto/tink/AeadStorageClient.kt#L39)):*
<details><summary><i>Previously, Marco-Premier (marcopremier) wrote…</i></summary><blockquote>

This wrapper statically declares `ConditionalOperationStorageClient` but only honors it if the wrapped client does, and we only find out via a per-call `require` that throws an undocumented,  `IllegalArgumentException`.

Would it worth to validate once at construction and hold a typed delegate so misconfiguration fails immediately instead of after compute.?

Same apply to `StreamingAeadStorageClient.kt` and `MesosRecordIoStorageClient.kt`

```
class AeadStorageClient(private val storageClient: StorageClient, private val aead: Aead) :

    StorageClient, ConditionalOperationStorageClient {

    private val conditionalDelegate: ConditionalOperationStorageClient =
      requireNotNull(storageClient as? ConditionalOperationStorageClient) {
        "AeadStorageClient requires a ConditionalOperationStorageClient, but wrapped " +
          "${storageClient::class.simpleName} does not support conditional writes"
      }   

      
    // then in the methods, drop the per-call require and use the typed delegate:
    override suspend fun writeBlobIfGeneration(
      blobKey: String,
      expectedGeneration: Long,
      content: Flow<ByteString>,
    ): StorageClient.Blob {

      val ciphertext: ByteString = /* ...encrypt... */

      val wrappedBlob = conditionalDelegate.writeBlobIfGeneration(blobKey, expectedGeneration, flow { emit(ciphertext) })
      return EncryptedBlob(wrappedBlob, blobKey)

    } 
```
</blockquote></details>

Done in b1f6b6c. Made it a hard requirement at construction across all three wrappers (`AeadStorageClient`, `StreamingAeadStorageClient`, `MesosRecordIoStorageClient`) — each now holds a typed `conditional: ConditionalOperationStorageClient` field validated by `requireNotNull(storageClient as? ConditionalOperationStorageClient) { ... }` at init. The per-call `require` blocks are gone (6 throws removed); construction-time fast-fail. Test cases updated accordingly to assert at construction rather than at the write call.
___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/tink/AeadStorageClient.kt` line 77 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/389#-OvJs_FxD_a_Ql8BfpqA:-OvKzo6SBfJ6O9nEaTII:b-ovj3oh) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/73cd7e1fc81dbe3765cef33789ffc54bf66f0848/src/main/kotlin/org/wfanet/measurement/common/crypto/tink/AeadStorageClient.kt#L77)):*
<details><summary><i>Previously, Marco-Premier (marcopremier) wrote…</i></summary><blockquote>

The `expectedGeneration >= 0` guard only exists in the leaf clients, shall we have early failure regardless of which underlying client is wrapped?

e.g.

```
override suspend fun writeBlobIfGeneration(
    blobKey: String,
    expectedGeneration: Long,
    content: Flow<ByteString>,
  ): StorageClient.Blob {

    require(expectedGeneration >= 0) { "expectedGeneration must be >= 0, got $expectedGeneration" }
    // ...existing body...
  }
```

Same apply to `StreamingAeadStorageClient.kt` and `MesosRecordIoStorageClient.kt`
</blockquote></details>

Done in b1f6b6c. `require(expectedGeneration >= 0)` now lives at the wrapper boundary in all three wrappers' `writeBlobIfGeneration`, so misuse fails locally regardless of which underlying client is wrapped.
___
*[`src/main/kotlin/org/wfanet/measurement/storage/MesosRecordIoStorageClient.kt` line 55 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/389#-OvJsDT06WcW7d8ZjGyk:-OvKznpW8BgJmsWaxZbE:b9gop4n) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/73cd7e1fc81dbe3765cef33789ffc54bf66f0848/src/main/kotlin/org/wfanet/measurement/storage/MesosRecordIoStorageClient.kt#L55)):*
<details><summary><i>Previously, Marco-Premier (marcopremier) wrote…</i></summary><blockquote>

The framing here is now duplicated by the new `recordIoFraming()` helper (line 117) — I'd suggest to have `writeBlob` reuse it so there's one source of truth for the size+delimiter prefix. 

```
override suspend fun writeBlob(blobKey: String, content: Flow<ByteString>): StorageClient.Blob {

    val wrappedBlob: StorageClient.Blob = storageClient.writeBlob(blobKey, recordIoFraming(content))
    return Blob(wrappedBlob, blobKey)

  } 
```
</blockquote></details>

Done in b1f6b6c. `writeBlob` now delegates to `storageClient.writeBlob(blobKey, recordIoFraming(content))` — single source of truth for the size+delimiter framing. The `recordsWritten` counter was lost in the inlining but it was only used for a `fine`-level log; the conditional variants don't log it either, so the loss is consistent.

stevenwarejones (commented):
> @stevenwarejones+AGNT made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/389)* status: 0 of 14 files reviewed, 3 unresolved discussions (waiting on Marco-Premier).

___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/tink/AeadStorageClient.kt` line 39 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/389#-OvJrWN-5WSwQspUpFvw:-OvL5XGzAgEaF5GPP0JU:b-udb83x) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/73cd7e1fc81dbe3765cef33789ffc54bf66f0848/src/main/kotlin/org/wfanet/measurement/common/crypto/tink/AeadStorageClient.kt#L39)):*
<details><summary><i>Previously, stevenwarejones+AGNT (Agent AGNT for Steven Ware Jones) wrote…</i></summary><blockquote>

Done in b1f6b6c. Made it a hard requirement at construction across all three wrappers (`AeadStorageClient`, `StreamingAeadStorageClient`, `MesosRecordIoStorageClient`) — each now holds a typed `conditional: ConditionalOperationStorageClient` field validated by `requireNotNull(storageClient as? ConditionalOperationStorageClient) { ... }` at init. The per-call `require` blocks are gone (6 throws removed); construction-time fast-fail. Test cases updated accordingly to assert at construction rather than at the write call.
</blockquote></details>

Follow-up in 7415403: tightened further, on review, to make the requirement compile-time rather than runtime. The wrapper constructors (`AeadStorageClient`, `StreamingAeadStorageClient`, `MesosRecordIoStorageClient`) now accept only `ConditionalOperationStorageClient` as the underlying client — the compiler catches misconfiguration at the wrapper construction site, no runtime path remains for "wrapped a non-conditional client." The typed `conditional` field and the `requireNotNull` cast are gone.

To keep the existing production callers compiling (notably `EncryptedStorage.buildEncryptedMesosStorageClient` through `SelectedStorageClient` which can return `FileSystemStorageClient`), `FileSystemStorageClient` now implements `ConditionalOperationStorageClient` too — generation is the file's `lastModified` (snapshotted at Blob construction), and conformance tests pass.

`WithEnvelopeEncryption.withEnvelopeEncryption` extension is now on `ConditionalOperationStorageClient`. All call sites compile unchanged because every concrete `StorageClient` in the codebase is now also `ConditionalOperationStorageClient`.

stevenwarejones (commented):
> @stevenwarejones reviewed 9 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/389)* status: 9 of 14 files reviewed, 3 unresolved discussions (waiting on Marco-Premier).

Marco-Premier (approved):
> @Marco-Premier reviewed 1 file and all commit messages, and resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/389)* status: 9 of 14 files reviewed, all discussions resolved.

stevenwarejones (commented):
> @stevenwarejones reviewed 10 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/389)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on SanjayVas).

SanjayVas (changes requested):
> @SanjayVas reviewed 1 file and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/389)* status: all files reviewed, 1 unresolved discussion (waiting on stevenwarejones).

___
*[`src/main/kotlin/org/wfanet/measurement/storage/StorageClient.kt` line 187 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/389#-Ovke8GD8gE8NV8jZZ92:-Ovke8GE6AaKVrMKFTdX:b-g1ra18) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/650eb238c018cdb641ef98aca43d2ebb6f05c5e6/src/main/kotlin/org/wfanet/measurement/storage/StorageClient.kt#L187)):*
> ```Kotlin
>    * @throws BlobChangedException if the precondition fails
>    */
>   suspend fun writeBlobIfGeneration(
> ```

Am I understanding correctly that the only reason this is needed is for the `0` case? The case of an existing blob should be covered by `writeBlobIfUnchanged`

If that's true, I think it may be better to have this explicitly be a method that conditionally writes the blob if it does not already exist (e.g. `writeBlobIfNotFound`) and rely on `writeBlobIfUnchanged` for the case where we know the Blob already exists.

The reasoning is that I'd rather avoid exposing the "generation" concept to callers and keep the mechanism for conditional writes as an implementation detail (e.g. generation for GCS, maybe an etag or write timestamp for other impls).

stevenwarejones (commented):
> @stevenwarejones made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/389)* status: all files reviewed, 1 unresolved discussion (waiting on SanjayVas).

___
*[`src/main/kotlin/org/wfanet/measurement/storage/StorageClient.kt` line 187 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/389#-Ovke8GD8gE8NV8jZZ92:-OvoaCcY3DOVLSeNK4oj:baurkpy) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/650eb238c018cdb641ef98aca43d2ebb6f05c5e6/src/main/kotlin/org/wfanet/measurement/storage/StorageClient.kt#L187)):*

Am I understanding correctly that the only reason this is needed is for the `0` case? The case of an existing blob should be covered by `writeBlobIfUnchanged`

If that's true, I think it may be better to have this explicitly be a method that conditionally writes the blob if it does not already exist (e.g. `writeBlobIfNotFound`) and rely on `writeBlobIfUnchanged` for the case where we know the Blob already exists.

The reasoning is that I'd rather avoid exposing the "generation" concept to callers and keep the mechanism for conditional writes as an implementation detail (e.g. generation for GCS, maybe an etag or write timestamp for other impls).
</blockquote></details>

That is not correct. I initially started to write it like you are thinking exactly but pivoted as we thought about it more. However, the use case in question is trying to be resilient against scenarios where the same work item is delivered to more than one MIG for a variety of reasons. There's a few layers of defense here but the idea is that the TEE Apps each grab the ifGeneration at the beginning of the workflow and write with that at the end to prevent one TEE from clobbering the work of the other, which in the case of VID memoization is potentially catastrophic and poisons VID labeling for every day after that clobbering.

SanjayVas (changes requested):
> @SanjayVas made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/389)* status: all files reviewed, 1 unresolved discussion (waiting on stevenwarejones).

___
*[`src/main/kotlin/org/wfanet/measurement/storage/StorageClient.kt` line 187 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/389#-Ovke8GD8gE8NV8jZZ92:-OvqmTFrDzW4sQXfTayK:b-e8uiec) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/650eb238c018cdb641ef98aca43d2ebb6f05c5e6/src/main/kotlin/org/wfanet/measurement/storage/StorageClient.kt#L187)):*

That is not correct. I initially started to write it like you are thinking exactly but pivoted as we thought about it more. However, the use case in question is trying to be resilient against scenarios where the same work item is delivered to more than one MIG for a variety of reasons. There's a few layers of defense here but the idea is that the TEE Apps each grab the ifGeneration at the beginning of the workflow and write with that at the end to prevent one TEE from clobbering the work of the other, which in the case of VID memoization is potentially catastrophic and poisons VID labeling for every day after that clobbering.
</blockquote></details>

If you need to persist some value for resource freshness validation outside of the blob, then it needs to be exposed from the blob in an opaque way. For example, a string "freshness token". There shouldn't be any transparent meaning to specific values such as `0`.

I think that there should still be a separate method to explicitly handle the case of not existing and that this one that takes in the opaque freshness token can just be an overload of `writeBlobIfUnchanged`.

stevenwarejones (commented):
> @stevenwarejones made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/389)* status: all files reviewed, 1 unresolved discussion (waiting on SanjayVas).

___
*[`src/main/kotlin/org/wfanet/measurement/storage/StorageClient.kt` line 187 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/389#-Ovke8GD8gE8NV8jZZ92:-OvtpGP1BPz5xfV6pNzw:b6twla) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/650eb238c018cdb641ef98aca43d2ebb6f05c5e6/src/main/kotlin/org/wfanet/measurement/storage/StorageClient.kt#L187)):*

If you need to persist some value for resource freshness validation outside of the blob, then it needs to be exposed from the blob in an opaque way. For example, a string "freshness token". There shouldn't be any transparent meaning to specific values such as `0`.

I think that there should still be a separate method to explicitly handle the case of not existing and that this one that takes in the opaque freshness token can just be an overload of `writeBlobIfUnchanged`.
</blockquote></details>

I mostly went with what you suggested using 
```
  suspend fun writeBlobIfUnchanged(
    blobKey: String,
    freshnessToken: String,
    content: Flow<ByteString>,
  ): Blob
```
and
```
  suspend fun writeBlobIfNotFound(blobKey: String, content: Flow<ByteString>): Blob
```
I added getFreshnessToken directly on the storage client rather than putting the token on Blob. The use case is a TEE that captures the version at workflow start and writes the result hours later, possibly from a different VM after Pub/Sub redelivery. A live Blob reference doesn't survive that boundary but a serializable String? token does. So the caller looks like:
```
// At workflow start (one VM):
  val token: String? = storage.getFreshnessToken(key)
  // ... hours of compute, possibly on a different VM after redelivery ...
  try {
    if (token == null) storage.writeBlobIfNotFound(key, content)
    else storage.writeBlobIfUnchanged(key, token, content)
  } catch (e: BlobChangedException) {
    // Another writer won; look up their row and ack.
  }
```

stevenwarejones (commented):
> @stevenwarejones reviewed 14 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/389)* status: all files reviewed, 1 unresolved discussion (waiting on SanjayVas).

SanjayVas (changes requested):
> @SanjayVas reviewed 4 files and all commit messages, made 2 comments, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/389)* status: all files reviewed, 2 unresolved discussions (waiting on stevenwarejones).

___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/tink/AeadStorageClient.kt` line 121 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/389#-Ow-6Ssp2EkNXShUB9X9:-Ow-6Ssp2EkNXShUB9XA:bipv7cw) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/62908f6f9024722efd93f08e66687ab42e0a7e56/src/main/kotlin/org/wfanet/measurement/common/crypto/tink/AeadStorageClient.kt#L121)):*
> ```Kotlin
>   /** A blob that will decrypt the content when read */
>   private inner class EncryptedBlob(
>     val underlying: StorageClient.Blob,
> ```

If you expose `freshnessToken` as a property of `EncryptedBlob`, then you shouldn't need to make `underlying` public.

My recommendation would be to add another Blob interface under `ConditionalOperationStorageClient` which defines this property and have the client-specific blob implementations implement that interface. You wouldn't have full type safety (for that we'd probably have to make `StorageClient` generic to enforce the Blob subtype), but we already have this problem where we need to cast the Blob.

You can still keep `getFreshnessToken` on `ConditionalOperationStorageClient` so callers don't need to do any casting.
___
*[`src/main/kotlin/org/wfanet/measurement/gcloud/gcs/GcsStorageClient.kt` line 153 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/389#-Ow-2nAz6lFZhUNFs8KT:-Ow-2nAz6lFZhUNFs8KU:b-wh3q5r) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/62908f6f9024722efd93f08e66687ab42e0a7e56/src/main/kotlin/org/wfanet/measurement/gcloud/gcs/GcsStorageClient.kt#L153)):*
> ```Kotlin
>     blobKey: String,
>     content: Flow<ByteString>,
>     coroutineName: String,
> ```

nit: I believe the idiomatic way to do this is to have the caller wrap this in `withContext(CoroutineName(coroutineName))` since contexts are inherited. You should generally not need to pass coroutine names outside of coroutine contexts.

stevenwarejones (commented):
> @stevenwarejones reviewed all commit messages, made 1 comment, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/389)* status: 8 of 15 files reviewed, 1 unresolved discussion (waiting on SanjayVas).

___
*[`src/main/kotlin/org/wfanet/measurement/common/crypto/tink/AeadStorageClient.kt` line 121 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/389#-Ow-6Ssp2EkNXShUB9X9:-Ow-K86b80Yi_qnQOTb5:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/62908f6f9024722efd93f08e66687ab42e0a7e56/src/main/kotlin/org/wfanet/measurement/common/crypto/tink/AeadStorageClient.kt#L121)):*

If you expose `freshnessToken` as a property of `EncryptedBlob`, then you shouldn't need to make `underlying` public.

My recommendation would be to add another Blob interface under `ConditionalOperationStorageClient` which defines this property and have the client-specific blob implementations implement that interface. You wouldn't have full type safety (for that we'd probably have to make `StorageClient` generic to enforce the Blob subtype), but we already have this problem where we need to cast the Blob.

You can still keep `getFreshnessToken` on `ConditionalOperationStorageClient` so callers don't need to do any casting.
</blockquote></details>

Done.

SanjayVas (approved):
> @SanjayVas reviewed 14 files and all commit messages, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/389)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on stevenwarejones).

stevenwarejones (commented):
> @stevenwarejones reviewed 7 files.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/389)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on stevenwarejones).

Marco-Premier (approved):
> @Marco-Premier reviewed 14 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/389)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on stevenwarejones).

PR #394, opened 2026-07-01
feat: add ParquetStorageClient.writeBlob overload that embeds footer key-value metadata

Marco-Premier:
## Summary

  `ParquetStorageClient` could already **read** footer key-value metadata via
  `ParquetBlob.readKeyValueMetadata()`, but had no way to **write** it. This adds the
  missing write counterpart: a `writeBlob` overload that accepts a
  `Map<String, String>` and embeds those entries into the parquet file's footer
  key-value metadata.

  ## Changes
  
  - New overload
    `suspend fun writeBlob(blobKey, content, keyValueMetadata: Map<String, String>)`
    that embeds the entries into the footer via parquet-mr's
    `ParquetWriter.Builder.withExtraMetaData` (available since parquet 1.12.0; the
    repo pins 1.14.3).
  - The existing `override writeBlob(blobKey, content)` now delegates to the new
    overload with `emptyMap()` — **no behavior change** for current callers.
  - The private `newWriter(...)` gained a `keyValueMetadata` parameter, threaded
    through **both** writer call sites (the main first-row-derived-schema path and
    the zero-row `EMPTY_SCHEMA` fallback), so empty-flow writes carry metadata too.
  - KDoc updated on the class and the new overload, including the PME
    `ENCRYPTED_FOOTER` caveat (see below).

  ## Notes / caveats
  
  - The footer is written on close, so metadata must be supplied at file creation —
    there is no in-place footer mutation.
  - Written entries are the inverse of `readKeyValueMetadata()`. When PME is
    configured in `ENCRYPTED_FOOTER` mode, the footer (and thus this metadata) is
    encrypted on disk and cannot be read back without the footer key; use
    `PLAINTEXT_FOOTER` (`parquet.encryption.plaintext.footer = true`) if the
    metadata must stay readable without the footer key.
  - No BUILD change required — `parquet_hadoop` is already a dependency.

  ## Testing
  
  Two new round-trip tests in `ParquetStorageClientTest.kt` (non-empty blob and
  empty/zero-row blob):

  //src/test/kotlin/org/wfanet/measurement/storage:ParquetStorageClientTest

Issue: world-federation-of-advertisers/cross-media-measurement#3948

### DIFF
--- a/src/main/kotlin/org/wfanet/measurement/storage/ParquetStorageClient.kt
+++ b/src/main/kotlin/org/wfanet/measurement/storage/ParquetStorageClient.kt
@@ -246,10 +246,9 @@ class ParquetStorageClient(
 
   /**
    * Encodes [content] (a flow of serialized [ParquetRow]) into a parquet file at [blobKey],
-   * optionally embedding [keyValueMetadata] into the file's footer key-value metadata (empty by
-   * default). The schema is derived from the first row; see the class KDoc for the first-row
-   * constraint. An empty flow writes a valid, zero-row parquet file, which still carries
-   * [keyValueMetadata].
+   * embedding [keyValueMetadata] into the file's footer key-value metadata. The schema is derived
+   * from the first row; see the class KDoc for the first-row constraint. An empty flow writes a
+   * valid, zero-row parquet file, which still carries [keyValueMetadata].
    *
    * The written entries are the inverse of [ParquetBlob.readKeyValueMetadata], which reads them
    * back. When PME is configured in `ENCRYPTED_FOOTER` mode the footer (and therefore this
@@ -260,7 +259,7 @@ class ParquetStorageClient(
   suspend fun writeBlob(
     blobKey: String,
     content: Flow<ByteString>,
-    keyValueMetadata: Map<String, String> = emptyMap(),
+    keyValueMetadata: Map<String, String>,
   ): StorageClient.Blob {
     val path = resolvePath(blobKey)
     withContext(parquetContext) {
--- a/src/main/kotlin/org/wfanet/measurement/storage/ParquetStorageClient.kt
+++ b/src/main/kotlin/org/wfanet/measurement/storage/ParquetStorageClient.kt
@@ -240,19 +240,16 @@ class ParquetStorageClient(
 
   // === StorageClient ===
 
-  /**
-   * Encodes [content] (a flow of serialized [ParquetRow]) into a parquet file at [blobKey]. The
-   * schema is derived from the first row; see the class KDoc for the first-row constraint. An empty
-   * flow writes a valid, zero-row parquet file.
-   */
+  /** Delegates to [writeBlob] with no footer key-value metadata. */
   override suspend fun writeBlob(blobKey: String, content: Flow<ByteString>): StorageClient.Blob =
     writeBlob(blobKey, content, emptyMap())
 
   /**
    * Encodes [content] (a flow of serialized [ParquetRow]) into a parquet file at [blobKey],
-   * embedding [keyValueMetadata] into the file's footer key-value metadata. The schema is derived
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/wfanet/measurement/storage/ParquetStorageClient.kt
```diff
@@ -229,12 +231,28 @@ class ParquetStorageClient(
...
+   * back. When PME is configured in `ENCRYPTED_FOOTER` mode the footer (and therefore this
+   * metadata) is encrypted on disk and [ParquetBlob.readKeyValueMetadata] cannot read it; write
+   * with `PLAINTEXT_FOOTER` (`parquet.encryption.plaintext.footer = true`) if the metadata must
+   * stay readable without the footer key.
    */
-  override suspend fun writeBlob(blobKey: String, content: Flow<ByteString>): StorageClient.Blob {
+  suspend fun writeBlob(
+    blobKey: String,
+    content: Flow<ByteString>,
+    keyValueMetadata: Map<String, String> = emptyMap(),
```
stevenwarejones: nit: Drop the ` = emptyMap()` default on `keyValueMetadata`. The overload above (line 235) already delegates the two-arg form to `writeBlob(blobKey, content, emptyMap())`, so this default is unreachable and just makes it look like there are two ways to spell the same call. Overloads-over-defaults is the WFA convention (Effective Java Item 52).

src/test/kotlin/org/wfanet/measurement/storage/ParquetStorageClientTest.kt
```diff
@@ -145,6 +145,30 @@ class ParquetStorageClientTest {
     assertThat(newClient().getBlob(key)!!.readKeyValueMetadata()).containsAtLeast("foo", "bar")
   }
 
+  @Test
+  fun `writeBlob embeds keyValueMetadata into the footer`(): Unit = runBlocking {
```
stevenwarejones, line 173: nit: Consider adding a round-trip test that combines the new overload with PME `PLAINTEXT_FOOTER` — write with `keyValueMetadata` through an encrypting client, then `readKeyValueMetadata` on a fresh non-encrypting client and assert the entries come back. The class KDoc calls out that these two features compose, and the existing `readKeyValueMetadata works on a plaintext-footer encrypted blob` (line 683) is close but uses the old no-metadata `writeBlob` and only asserts `isNotNull`. One extra test pins down the guarantee.

stevenwarejones (approved):
LGTM once the missing `Issue:` trailer is added to the PR body (blocking-on-CI). Two nits inline that are non-blocking.

stevenwarejones (approved):
> @stevenwarejones reviewed 2 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/394)* status: all files reviewed, 2 unresolved discussions (waiting on Marco-Premier).

Marco-Premier (commented):
> @Marco-Premier resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/394)* status: 0 of 2 files reviewed, all discussions resolved (waiting on stevenwarejones).

stevenwarejones (approved):
> @stevenwarejones reviewed 2 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/394)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on Marco-Premier).

SanjayVas (approved):
> @SanjayVas reviewed 2 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/394)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on Marco-Premier).

PR #395, opened 2026-07-02
feat(storage): expose ParquetBlob.readSchema()

Marco-Premier:
## Summary

Adds `ParquetBlob.readSchema(): Map<String, ParquetValue.KindCase>` — the file's column name → the `ParquetValue` kind that `readRows()` yields for it (same parquet-type → kind mapping already documented on `readRows`), reading only the schema (empty map for a zero-row file).

## Motivation

Downstream consumers map config fields to Parquet **column names** but only ever see per-row value maps, so a config that references a non-existent/renamed column is indistinguishable from a null cell and fails silently (the field is left unset). Exposing the schema lets a caller validate up front that every referenced column exists with the expected type and fail loud, instead of discovering drift at row time.

## Changes

- `ParquetBlob.readSchema()` on the interface + `ParquetBlobImpl`.
- Private `kindOf(Type)` mirroring `buildPrimitiveExtractor`'s primitive/logical-type dispatch, so the declared kinds match what a row read would produce.
- Tests: column-name→kind mapping; empty (zero-row) file returns empty.

### DIFF
--- a/src/main/kotlin/org/wfanet/measurement/storage/ParquetStorageClient.kt
+++ b/src/main/kotlin/org/wfanet/measurement/storage/ParquetStorageClient.kt
@@ -225,6 +225,15 @@ class ParquetStorageClient(
      * Throws [IllegalStateException] if the file is a PME `ENCRYPTED_FOOTER` blob (not supported).
      */
     suspend fun readKeyValueMetadata(): Map<String, String>
+
+    /**
+     * The file's parquet schema as a map of column name to the [ParquetValue.KindCase] that
+     * [readRows] yields for that column (same parquet-type -> kind mapping documented on
+     * [readRows]). Reads only the schema, not row values; returns an empty map for a zero-row file.
+     * Lets a caller validate up front that a config's referenced columns exist with the expected
+     * types, instead of discovering a missing/renamed column silently at row time.
+     */
+    suspend fun readSchema(): Map<String, ParquetValue.KindCase>
   }
 
   // === StorageClient ===
@@ -373,6 +382,15 @@ class ParquetStorageClient(
     override suspend fun readKeyValueMetadata(): Map<String, String> =
       withContext(parquetContext) { readFooterKeyValueMetadata(path) }
 
+    override suspend fun readSchema(): Map<String, ParquetValue.KindCase> =
+      withContext(parquetContext) {
+        val inputFile: InputFile = HadoopInputFile.fromPath(path, conf)
+        GroupParquetReaderBuilder(inputFile).withConf(conf).build().use { reader ->
+          val first = reader.read() ?: return@use emptyMap<String, ParquetValue.KindCase>()
+          first.type.fields.associate { field -> field.name to kindOf(field) }
+        }
+      }
+
     override suspend fun delete() {
       withContext(parquetContext) { fileSystem.delete(path, /* recursive= */ false) }
     }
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/wfanet/measurement/storage/ParquetStorageClient.kt
```diff
@@ -373,6 +382,15 @@ class ParquetStorageClient(
     override suspend fun readKeyValueMetadata(): Map<String, String> =
       withContext(parquetContext) { readFooterKeyValueMetadata(path) }
 
+    override suspend fun readSchema(): Map<String, ParquetValue.KindCase> =
```
stevenwarejones, line 405: blocking: `readSchema` reads a full row to get the schema, and returns an empty map on a zero-row file — both wrong.

`GroupParquetReaderBuilder(inputFile).withConf(conf).build().use { reader -> val first = reader.read() ?: return@use emptyMap(); first.type.fields... }` triggers row-group decompression, per-column decoding, and materializes a `Group`, when the schema is available directly from the footer's `MessageType` with no row-group access. A zero-row file HAS a schema (Parquet stores it in the footer regardless of row count) — returning an empty map here means every caller doing "validate config columns against schema" silently no-ops on empty files. That's precisely the drift-detection failure the PR motivation cites.

Second problem: `ParquetFileReader.open` (invoked by the reader builder) decrypts ALL column metadata eagerly under PME `PLAINTEXT_FOOTER` mode — needs the footer key even though the schema itself is plaintext. `readFooterKeyValueMetadata` at line 738 goes to real lengths to avoid this exact pitfall via `Util.readFileMetaData(skipRowGroups = true)`; `readSchema` should follow the same pattern.

Rewrite to reuse the direct-thrift footer path:

```kotlin
override suspend fun readSchema(): Map<String, ParquetValue.KindCase> =
  withContext(parquetContext) { readFooterSchema(path) }

private fun readFooterSchema(path: Path): Map<String, ParquetValue.KindCase> {
  // Same trailer parse as readFooterKeyValueMetadata (extract into a shared helper).
  // After parsing the footer bytes:
  val md = Util.readFileMetaData(ByteArrayInputStream(footerBytes), /* skipRowGroups= */ true)
  val messageType = ParquetMetadataConverter().fromParquetSchema(md.schema)
  return messageType.fields.associate { it.name to kindOf(it) }
}
```

Update the docstring: drop the "empty map for zero-row file" clause; every valid Parquet file has a schema, and `readSchema` should reflect it. The test `readSchema returns empty for a zero-row file` is asserting the wrong contract and should be replaced with `readSchema returns declared schema for a zero-row file`.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/395)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on Marco-Premier and stevenwarejones).

stevenwarejones (commented):
> @stevenwarejones reviewed 2 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/395)* status: all files reviewed, 1 unresolved discussion (waiting on Marco-Premier).

src/main/kotlin/org/wfanet/measurement/storage/ParquetStorageClient.kt
Marco-Premier: Done.

Marco-Premier (commented):
I have accidentally merged the already-approved #394  into this branch, sorry about that. can I not-undo the merge?

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/395#-:-Owvat8j1KEP9oFkwBWq:bxxhesx)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on stevenwarejones and stevenwarejones+AGNT).

src/main/kotlin/org/wfanet/measurement/storage/ParquetStorageClient.kt
stevenwarejones: Verified at head 7b44b8be — readSchema now routes through the direct-thrift footer parse via a new readFooterFileMetaData helper (extracted from readFooterKeyValueMetadata), converts via ParquetMetadataConverter().fromParquetMetadata(md).fileMetaData.schema, and maps through the existing kindOf. Docstring updated. Test replaced with readSchema returns declared schema for a zero-row file asserting the actual contract. All three parts of the blocking landed exactly as drafted. Fixed.

stevenwarejones (commented):
> @stevenwarejones+AGNT made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/395)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on stevenwarejones).

stevenwarejones (approved):
> @stevenwarejones reviewed 2 files and all commit messages, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/395)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on Marco-Premier).

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/395)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on Marco-Premier).

SanjayVas (approved):
> @SanjayVas reviewed 2 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/395)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on Marco-Premier).

PR #396, opened 2026-07-03
build: Update dependencies

laureanobrs:
This PR
- updates the version of protobuf, gRPC, and a few other dependencies.
- stops marking some maven artifacts as test only.


laureanobrs (commented):
> @laureanobrs made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/396)* status: 0 of 3 files reviewed, all discussions resolved.

___
*[`MODULE.bazel` line 98 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/396#-OxC9vF76a-Sdj3ZDwrj:-OxC9vF8DGDnWxO8tkbV:b-a4s12y) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/7993f6634762a0f4ecddf51b0ddbb686192a7a65/MODULE.bazel#L98)):*
> ```Starlark
> maven = use_extension("@rules_jvm_external//:extensions.bzl", "maven")
> maven.artifact(
>     testonly = True,
> ```

This PR also stops marking these artifacts as `testonly`:

```text id="m7p4x9"
artifact = "mockito-core"
artifact = "mockito-kotlin"
artifact = "grpc-testing"
```

They are required by this target:
https://github.com/world-federation-of-advertisers/common-jvm/blob/4c586851c54f9b18e637df43c4303b8613b18456/src/main/kotlin/org/wfanet/measurement/common/grpc/testing/BUILD.bazel#L16

Which is used in CMM. This bug becomes visible when updating the `rules_jvm_external` version in CMM.

laureanobrs (commented):
> @laureanobrs reviewed 2 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/396)* status: 2 of 3 files reviewed, all discussions resolved.

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/396#-OxWjm2W95HhCwfUaP3_:-OxWjm2W95HhCwfUaP3a:b-7pln7e) (no related file):*
This PR also stops marking some artifacts as `testonly` because they are required by this target:

https://github.com/world-federation-of-advertisers/common-jvm/blob/4c586851c54f9b18e637df43c4303b8613b18456/src/main/kotlin/org/wfanet/measurement/common/grpc/testing/BUILD.bazel#L16

Which is public and used in CMM. This bug becomes visible when updating the `rules_jvm_external` version in CMM to >6.8.

SanjayVas (changes requested):
> @SanjayVas made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/396)* status: 2 of 3 files reviewed, 1 unresolved discussion (waiting on laureanobrs).

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/396#-OxWjm2W95HhCwfUaP3_:-OxXXlCJCHZCj2RnJLK3:b7f6bsa) (no related file):*
<details><summary><i>Previously, laureanobrs (Laureano Ceschini) wrote…</i></summary><blockquote>

This PR also stops marking some artifacts as `testonly` because they are required by this target:

https://github.com/world-federation-of-advertisers/common-jvm/blob/4c586851c54f9b18e637df43c4303b8613b18456/src/main/kotlin/org/wfanet/measurement/common/grpc/testing/BUILD.bazel#L16

Which is public and used in CMM. This bug becomes visible when updating the `rules_jvm_external` version in CMM to >6.8.
</blockquote></details>

That target is itself marked as testonly due to the `default_testonly` of the package. This would not prevent the artifact from being marked as testonly.

laureanobrs (commented):
> @laureanobrs made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/396)* status: 1 of 3 files reviewed, 1 unresolved discussion (waiting on SanjayVas).

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/396#-OxWjm2W95HhCwfUaP3_:-OxbCg73A6G8_32DQbOK:blzipz0) (no related file):*

That target is itself marked as testonly due to the `default_testonly` of the package. This would not prevent the artifact from being marked as testonly.
</blockquote></details>

Regarding all the cases where I removed `testonly`, I did so because they were all causing errors like this during the build of CMM when using the latest version of `rules_jvm_external`:

```text
ERROR: Analysis of target '//src/main/kotlin/org/wfanet/measurement/integration/common:in_process_duchy' failed; build aborted: Analysis failed

ERROR: /home/runner/.cache/bazel/_bazel_runner/5efb41e3bf38a44ce0efecb45b2f3782/external/common-jvm+/imports/java/io/grpc/testing/BUILD.bazel:3:6:
    no such target '@@rules_jvm_external++maven+maven//:io_grpc_grpc_testing':
    target 'io_grpc_grpc_testing' not declared in package ''
    defined by /home/runner/.cache/bazel/_bazel_runner/5efb41e3bf38a44ce0efecb45b2f3782/external/rules_jvm_external++maven+maven/BUILD
    and referenced by '@@common-jvm+//imports/java/io/grpc/testing:testing'
```

I reproduced one of those cases by adding this [temporary commit to common-jvm](https://github.com/world-federation-of-advertisers/common-jvm/pull/396/changes/a3eaafdeb31ce4fd28acf7b8e0d177182f2e3a8d), defining `io_grpc_grpc_testing` as `testonly`. 
I also added this [temporary commit to CMM](https://github.com/world-federation-of-advertisers/cross-media-measurement/pull/4189/changes/ad1085ac838a57069c6c3c176af746f56c617afa), using that `common-jvm` commit and version 7.0 of `rules_jvm_external`. And also set `fetch_sources = false` as a workaround for the `rules_jvm_external` issue that led me to add the `single_version_override`.

Here is the link to the failed build:

[https://github.com/world-federation-of-advertisers/cross-media-measurement/actions/runs/29442979118/job/87446315247?pr=4189#step:9:79](https://github.com/world-federation-of-advertisers/cross-media-measurement/actions/runs/29442979118/job/87446315247?pr=4189#step:9:79)

* * *

Although I did not update the version of `rules_jvm_external` in the CMM PR—which is what triggers the error related to `testonly` dependencies—I already included the necessary changes here (and in rules_kotlin_jvm) so that the repository will be ready to update `rules_jvm_external` once the fix becomes available. This avoids having to redo work I have already completed.

SanjayVas (changes requested):
> @SanjayVas made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/396)* status: 1 of 3 files reviewed, 1 unresolved discussion (waiting on laureanobrs).

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/396#-OxWjm2W95HhCwfUaP3_:-OxbkObF1Qgy9QDOYJRf:b-tgxpm5) (no related file):*
> This avoids having to redo work I have already completed.

I don't understand. If there's a bug that's causing us to need to stop marking these as testonly, we'd need a TODO to restore our code to the state it would be in if the bug were not present. Therefore you'd still have work to do once the fix becomes available.

laureanobrs (commented):
> @laureanobrs made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/396)* status: 1 of 3 files reviewed, 1 unresolved discussion (waiting on SanjayVas).

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/396#-OxWjm2W95HhCwfUaP3_:-Oxbnj8T3BRNIx1mT8ps:b-brtvq8) (no related file):*

> This avoids having to redo work I have already completed.

I don't understand. If there's a bug that's causing us to need to stop marking these as testonly, we'd need a TODO to restore our code to the state it would be in if the bug were not present. Therefore you'd still have work to do once the fix becomes available.
</blockquote></details>

What I meant with avoiding to having to redo work is:

The `testonly` bug occurs with `rules_jvm_external` versions later than 6.8 in the CMM repository. Since CMM will continue using version 6.8 until an unrelated issue is fixed, there is no immediate need to include these `testonly` changes.

I included them now to avoid having to identify again which libraries should no longer be marked as `testonly` once we eventually upgrade.

SanjayVas (changes requested):
> @SanjayVas made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/396)* status: 1 of 3 files reviewed, 1 unresolved discussion (waiting on laureanobrs).

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/396#-OxWjm2W95HhCwfUaP3_:-Oxc2uJV5ku0TKmVu1MA:budflht) (no related file):*
<details><summary><i>Previously, laureanobrs (Laureano Ceschini) wrote…</i></summary><blockquote>

What I meant with avoiding to having to redo work is:

The `testonly` bug occurs with `rules_jvm_external` versions later than 6.8 in the CMM repository. Since CMM will continue using version 6.8 until an unrelated issue is fixed, there is no immediate need to include these `testonly` changes.

I included them now to avoid having to identify again which libraries should no longer be marked as `testonly` once we eventually upgrade.
</blockquote></details>

These targets *should be* testonly and we *want* them that way. If they cannot be, the reason for changing that would need to be clearly investigated and documented. This means figuring out exactly what changed in `rules_jvm_external` and whether that change is intentional/correct or a bug.

I would be suspicious of it being correct as I'd expect these targets to only be used for testing and thus safe to mark as `testonly`. I believe the only dep that has caused us to mark test deps as not testonly was Apache Beam. I think this was due to mixing testing and non-testing code in the same Maven artifact. That said, we may be doing the same for Maven artifacts from common-jvm. It may be that we need to fix that.

laureanobrs (commented):
> @laureanobrs made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/396)* status: 1 of 3 files reviewed, 1 unresolved discussion (waiting on SanjayVas).

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/396#-OxWjm2W95HhCwfUaP3_:-OxfFwxk23gBg7t4X-Vn:bodwxxi) (no related file):*

These targets *should be* testonly and we *want* them that way. If they cannot be, the reason for changing that would need to be clearly investigated and documented. This means figuring out exactly what changed in `rules_jvm_external` and whether that change is intentional/correct or a bug.

I would be suspicious of it being correct as I'd expect these targets to only be used for testing and thus safe to mark as `testonly`. I believe the only dep that has caused us to mark test deps as not testonly was Apache Beam. I think this was due to mixing testing and non-testing code in the same Maven artifact. That said, we may be doing the same for Maven artifacts from common-jvm. It may be that we need to fix that.
</blockquote></details>

To clarify what I meant, whenever I referred to this as a “bug,” it was because I had Maven’s model for test-only dependencies in mind, which is non-transitive, whereas we have been treating it as transitive here.

I had assumed that Maven’s `test` scope should match the behavior of Bazel’s `dev_dependency` attribute rather than that of `testonly`.

Separately, this is the `rules_jvm_external` PR that, as I understand it, changed RJE’s `test_only` behavior so that it is no longer transitive:

[https://github.com/bazel-contrib/rules\_jvm\_external/pull/1511/changes#diff-098382e4441ec77ea0bc64b27d5a3e7d87faa8eb632ce50c2abfab57553373c0](https://github.com/bazel-contrib/rules_jvm_external/pull/1511/changes#diff-098382e4441ec77ea0bc64b27d5a3e7d87faa8eb632ce50c2abfab57553373c0)

The change was included in the 6.10 release:

[https://github.com/bazel-contrib/rules\_jvm\_external/releases#release-6.10](https://github.com/bazel-contrib/rules_jvm_external/releases#release-6.10)

SanjayVas (changes requested):
> @SanjayVas made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/396)* status: 1 of 3 files reviewed, 1 unresolved discussion (waiting on laureanobrs).

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/396#-OxWjm2W95HhCwfUaP3_:-OxfvIJh2ldCWkipjAWH:bc2h8gs) (no related file):*
<details><summary><i>Previously, laureanobrs (Laureano Ceschini) wrote…</i></summary><blockquote>

To clarify what I meant, whenever I referred to this as a “bug,” it was because I had Maven’s model for test-only dependencies in mind, which is non-transitive, whereas we have been treating it as transitive here.

I had assumed that Maven’s `test` scope should match the behavior of Bazel’s `dev_dependency` attribute rather than that of `testonly`.

Separately, this is the `rules_jvm_external` PR that, as I understand it, changed RJE’s `test_only` behavior so that it is no longer transitive:

[https://github.com/bazel-contrib/rules\_jvm\_external/pull/1511/changes#diff-098382e4441ec77ea0bc64b27d5a3e7d87faa8eb632ce50c2abfab57553373c0](https://github.com/bazel-contrib/rules_jvm_external/pull/1511/changes#diff-098382e4441ec77ea0bc64b27d5a3e7d87faa8eb632ce50c2abfab57553373c0)

The change was included in the 6.10 release:

[https://github.com/bazel-contrib/rules\_jvm\_external/releases#release-6.10](https://github.com/bazel-contrib/rules_jvm_external/releases#release-6.10)
</blockquote></details>

So are you saying that the rules_jvm_external change made it so that the `testonly` attribute is not applied for artifacts from non-root modules? That should be filed as a bug. Even if they did it intentionally, that's a breaking change.

laureanobrs (commented):
> @laureanobrs made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/396)* status: 1 of 3 files reviewed, 1 unresolved discussion (waiting on SanjayVas).

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/396#-OxWjm2W95HhCwfUaP3_:-OxfyzY16rzS-lbo4oVZ:brmsrgg) (no related file):*

So are you saying that the rules_jvm_external change made it so that the `testonly` attribute is not applied for artifacts from non-root modules? That should be filed as a bug. Even if they did it intentionally, that's a breaking change.
</blockquote></details>

I'm saying that the `rules_jvm_external` change made the `testonly` attribute cause those artifacts to be excluded when contributor modules are merged into the root module.

That’s what I’m seeing in practice, and it also matches how I interpret the first half of [that PR title](https://github.com/bazel-contrib/rules\_jvm\_external/pull/1511/changes#diff-098382e4441ec77ea0bc64b27d5a3e7d87faa8eb632ce50c2abfab57553373c0): “Filter `test_only` artifacts out of artifacts merged into root repos and …”

SanjayVas (changes requested):
> @SanjayVas made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/396)* status: 1 of 3 files reviewed, 1 unresolved discussion (waiting on laureanobrs).

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/396#-OxWjm2W95HhCwfUaP3_:-OxgBB0S2eGvqojVVv75:bsh25wj) (no related file):*
> I'm saying that the `rules_jvm_external` change made the `testonly` attribute cause those artifacts to be excluded when contributor modules are merged into the root module.

That's still a breaking change that they shouldn't be pushing in a minor release. Maven testonly and Bazel testonly are indeed being conflated.

laureanobrs (commented):
> @laureanobrs made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/396)* status: 1 of 3 files reviewed, 1 unresolved discussion (waiting on SanjayVas).

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/396#-OxWjm2W95HhCwfUaP3_:-OxgIQJY5CkQuuRDJMq_:bdsp7th) (no related file):*

> I'm saying that the `rules_jvm_external` change made the `testonly` attribute cause those artifacts to be excluded when contributor modules are merged into the root module.

That's still a breaking change that they shouldn't be pushing in a minor release. Maven testonly and Bazel testonly are indeed being conflated.
</blockquote></details>

I agree with that. What do you think we should do about it?

I don’t feel I have a strong enough argument to file an issue against RJE claiming that they should not have introduced this breaking change. I haven’t found anything authoritative that I can cite to argue that they should not follow Maven’s handling of the `test` scope, which is non-transitive, and should instead preserve Bazel’s `testonly` semantics, which are transitive.

SanjayVas (changes requested):
> @SanjayVas made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/396)* status: 1 of 3 files reviewed, 1 unresolved discussion (waiting on laureanobrs).

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/396#-OxWjm2W95HhCwfUaP3_:-OxgKS6HDrwCBc0fPDDz:bovyk0y) (no related file):*
<details><summary><i>Previously, laureanobrs (Laureano Ceschini) wrote…</i></summary><blockquote>

I agree with that. What do you think we should do about it?

I don’t feel I have a strong enough argument to file an issue against RJE claiming that they should not have introduced this breaking change. I haven’t found anything authoritative that I can cite to argue that they should not follow Maven’s handling of the `test` scope, which is non-transitive, and should instead preserve Bazel’s `testonly` semantics, which are transitive.
</blockquote></details>

It's pretty clear to me. Filed https://github.com/bazel-contrib/rules_jvm_external/issues/1609

PR #398, opened 2026-07-23
feat: resilient GCS retry/timeout defaults for transient reads

Marco-Premier:
## Why
  
  Cloud Storage reads in downstream services (observed in the EDPA Results-Fulfiller)
  intermittently fail on transient connection resets — `com.google.cloud.storage.StorageException:
  Broken pipe` and `Remote host terminated the handshake`, thrown from
  `BaseStorageReadChannel.read`. Root cause is client configuration, not infrastructure:
  
  - `StorageOptions` was built with **no transport socket read timeout** and the
    google-cloud-storage **library-default retry budget** (`totalTimeout = 50s`,
    `maxAttempts = 6`).
  - Production `RetryContext$RetryBudgetExhausted` traces show reads failing after
    **`attempts: 1`**: a single stalled/slow socket read consumes the whole 50s budget before
    a second attempt can start, because nothing bounds an individual stalled read.
  - The google-cloud-storage read channel **already resumes mid-stream at the current offset**
    via a `Range` request (`ApiaryUnbufferedReadableByteChannel` tracks `position` and re-opens
    with `withNewBeginOffset`) — but with the default budget it never gets to run. 
    
  ## What bounds a stalled read (the key knob)
  
  On the HTTP/Apiary transport the read is aborted by the **transport socket read timeout**
  (`HttpTransportOptions.setReadTimeout` → `HttpRequest.setReadTimeout` → `HttpURLConnection`
  `SO_TIMEOUT`), applied via `StorageOptions.setTransportOptions` — **not** by the
  `RetrySettings` RPC timeouts, which are not enforced on the blocking Apiary read.
  `RetrySettings` governs only the retry budget (`totalTimeout`, `maxAttempts`) and backoff.

  ## Changes

  - New `GcsStorageOptions.kt`: a validated `GcsStorageRetryConfig` and a single
    `buildGcsStorageOptions(...)` factory — the one place `StorageOptions` is now built.
  - Resilient **defaults**, deliberately generous so legitimate slow reads are not broken:
    - per-attempt `readTimeout = 30s`, `connectTimeout = 15s` (HTTP transport);
    - `totalTimeout = 180s` (was 50s) so **several** bounded attempts fit, each resuming at
      the current offset; `maxAttempts = 6`, jittered exponential backoff (1s → 32s).
    - Fully overridable via the optional `retryConfig` parameter, so a caller (e.g. the
      Results-Fulfiller) can tune tighter.
  - `SelectedStorageClient` and `GcsFromFlags` route through the factory. All public APIs stay
    **backward-compatible** (new parameters are optional/defaulted).

  ## Testing

  - **Reproduces the actual glitch:** a raw `ServerSocket` sends 16 KiB then a TCP RST
    mid-download; the test asserts the full blob reads back byte-identical **and**
    `observedMediaRangeStarts == [0, 16384]` — i.e. the client genuinely resumes via
    `Range: bytes=16384-` with no double-count.
  - **Stalled-read test:** first attempt stalls; the per-attempt read timeout aborts it and the
    retry succeeds (fast, not after the full stall).
  - Config-mapping/validation unit tests and a backward-compat test for a custom `retryConfig`.
  - Existing `SelectedStorageClientTest` and `GcsStorageClientTest` pass (no regression).
  - `ktfmt 0.54 --google-style` clean.

Issue: https://github.com/world-federation-of-advertisers/cross-media-measurement/issues/4286

### DIFF
--- a/src/main/kotlin/org/wfanet/measurement/gcloud/gcs/BUILD.bazel
+++ b/src/main/kotlin/org/wfanet/measurement/gcloud/gcs/BUILD.bazel
@@ -8,7 +8,9 @@ kt_jvm_library(
     name = "gcs",
     srcs = glob(["*.kt"]),
     deps = [
+        "//imports/java/com/google/api/gax",
         "//imports/java/com/google/cloud/storage",
+        "//imports/java/io/opentelemetry/api",
         "//imports/java/picocli",
         "//imports/kotlin/kotlinx/coroutines:core",
         "//src/main/kotlin/org/wfanet/measurement/common",
--- a/src/main/kotlin/org/wfanet/measurement/gcloud/gcs/GcsFromFlags.kt
+++ b/src/main/kotlin/org/wfanet/measurement/gcloud/gcs/GcsFromFlags.kt
@@ -19,18 +19,25 @@ import com.google.cloud.storage.StorageOptions
 import org.wfanet.measurement.common.Instrumentation
 import picocli.CommandLine
 
-/** Client access provider for Google Cloud Storage (GCS) via command-line flags. */
-class GcsFromFlags(private val flags: Flags) {
+/**
+ * Client access provider for Google Cloud Storage (GCS) via command-line flags.
+ *
+ * @param flags command-line flags
+ * @param retryConfig resilient retry/timeout configuration applied to the GCS client; defaults to
+ *   [GcsStorageRetryConfig.DEFAULT]
+ */
+class GcsFromFlags(
+  private val flags: Flags,
+  private val retryConfig: GcsStorageRetryConfig = GcsStorageRetryConfig.DEFAULT,
+) {
 
   private val storageOptions: StorageOptions by lazy {
-    val builder =
-      if (flags.useGrpc) {
-        StorageOptions.grpc().setEnableGrpcClientMetrics(true)
-      } else {
-        StorageOptions.newBuilder()
-      }
-
-    builder.setOpenTelemetry(Instrumentation.openTelemetry).setProjectId(flags.projectName).build()
+    buildGcsStorageOptions(
+      projectId = flags.projectName,
+      useGrpc = flags.useGrpc,
+      openTelemetry = Instrumentation.openTelemetry,
+      retryConfig = retryConfig,
+    )
   }
 
   val storage: Storage
--- a/src/main/kotlin/org/wfanet/measurement/gcloud/gcs/GcsFromFlags.kt
+++ b/src/main/kotlin/org/wfanet/measurement/gcloud/gcs/GcsFromFlags.kt
@@ -32,11 +32,10 @@ class GcsFromFlags(
 ) {
 
   private val storageOptions: StorageOptions by lazy {
-    buildGcsStorageOptions(
+    retryConfig.buildStorageOptions(
       projectId = flags.projectName,
       useGrpc = flags.useGrpc,
       openTelemetry = Instrumentation.openTelemetry,
-      retryConfig = retryConfig,
     )
   }
 
--- a/src/main/kotlin/org/wfanet/measurement/gcloud/gcs/GcsStorageOptions.kt
+++ b/src/main/kotlin/org/wfanet/measurement/gcloud/gcs/GcsStorageOptions.kt
@@ -39,13 +39,22 @@ import java.time.Duration
  *   ([com.google.cloud.storage.ApiaryUnbufferedReadableByteChannel], which tracks `position` and
  *   re-opens with `withNewBeginOffset`).
  *
- * ## What bounds a stalled read on the HTTP/Apiary transport
- * The knob that aborts a stalled socket read is the transport read timeout
- * ([HttpTransportOptions.Builder.setReadTimeout] -> `HttpRequest.setReadTimeout` ->
- * `HttpURLConnection` `SO_TIMEOUT`), applied via [StorageOptions.Builder.setTransportOptions]. This
- * is distinct from [RetrySettings] RPC timeouts (`initialRpcTimeout`/`maxRpcTimeout`), which are
- * not enforced on the blocking Apiary HTTP read. [RetrySettings] here governs the retry budget
- * (`totalTimeout`, `maxAttempts`) and the jittered exponential backoff between attempts.
+ * ## What bounds an attempt (HTTP vs gRPC)
+ * The two transports bound a single attempt differently:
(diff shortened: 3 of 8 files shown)


SanjayVas (changes requested):
> @SanjayVas reviewed 8 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/398)* status: all files reviewed, 2 unresolved discussions (waiting on Marco-Premier).

___
*[`src/main/kotlin/org/wfanet/measurement/gcloud/gcs/GcsStorageOptions.kt` line 26 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/398#-OyF3ERiBLvasBvgQ7r-:-OyF3ERiBLvasBvgQ7r0:bfhbdak) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/03329c61905f6a588c6e0727b743aeb52cdb61c9/src/main/kotlin/org/wfanet/measurement/gcloud/gcs/GcsStorageOptions.kt#L26)):*
> ```Kotlin
>  * Resilience configuration applied to every [StorageOptions] built by [buildGcsStorageOptions].
>  *
>  * ## Why this exists
> ```

We don't need all of this context in code. Recall that comments/documentation in code have the same requirements as code itself, in that it is read many more times than it is written and it must be maintained. This means that every time someone looks at this code, they need to read through all this documentation. Comments are additionally more difficult to maintain due to the fact that tooling frequently has less understanding of comments. Therefore comments should be concise in the same way code should be concise.

Also make sure to put explanations in the right place: Issue investigation and explanation into the GitHub issue, a short description of the PR into the PR description, and appropriate documentation within the code. There should be little to no duplication of content between these locations.
___
*[`src/main/kotlin/org/wfanet/measurement/gcloud/gcs/GcsStorageOptions.kt` line 178 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/398#-OyF1XQtAKPsFo3Mhp4U:-OyF1XQuALg1XBoGKW9g:b884xmu) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/03329c61905f6a588c6e0727b743aeb52cdb61c9/src/main/kotlin/org/wfanet/measurement/gcloud/gcs/GcsStorageOptions.kt#L178)):*
> ```Kotlin
>  * @param retryConfig resilience configuration; defaults to [GcsStorageRetryConfig.DEFAULT].
>  */
> fun buildGcsStorageOptions(
> ```

Don't add new public top-level non-extension functions.

A couple options:

1. Create a GcsStorageOptions wrapper object.
2. Add this as a `buildStorageOptions` method on `GcsStorageRetryConfig` where it uses the corresponding instance rather than having a `retryConfig` param.

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/398)* status: 4 of 8 files reviewed, 2 unresolved discussions (waiting on SanjayVas).

___
*[`src/main/kotlin/org/wfanet/measurement/gcloud/gcs/GcsStorageOptions.kt` line 26 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/398#-OyF3ERiBLvasBvgQ7r-:-OyI0ACM5dINfPF0PxFz:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/03329c61905f6a588c6e0727b743aeb52cdb61c9/src/main/kotlin/org/wfanet/measurement/gcloud/gcs/GcsStorageOptions.kt#L26)):*

We don't need all of this context in code. Recall that comments/documentation in code have the same requirements as code itself, in that it is read many more times than it is written and it must be maintained. This means that every time someone looks at this code, they need to read through all this documentation. Comments are additionally more difficult to maintain due to the fact that tooling frequently has less understanding of comments. Therefore comments should be concise in the same way code should be concise.

Also make sure to put explanations in the right place: Issue investigation and explanation into the GitHub issue, a short description of the PR into the PR description, and appropriate documentation within the code. There should be little to no duplication of content between these locations.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/gcloud/gcs/GcsStorageOptions.kt` line 178 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/398#-OyF1XQtAKPsFo3Mhp4U:-OyI09Xu-ZPDcwVZIBhp:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/03329c61905f6a588c6e0727b743aeb52cdb61c9/src/main/kotlin/org/wfanet/measurement/gcloud/gcs/GcsStorageOptions.kt#L178)):*

Don't add new public top-level non-extension functions.

A couple options:

1. Create a GcsStorageOptions wrapper object.
2. Add this as a `buildStorageOptions` method on `GcsStorageRetryConfig` where it uses the corresponding instance rather than having a `retryConfig` param.
</blockquote></details>

Done.

SanjayVas (approved):
> @SanjayVas reviewed 4 files and all commit messages, made 1 comment, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/398)* status: all files reviewed, 1 unresolved discussion (waiting on Marco-Premier).

___
*[`src/main/kotlin/org/wfanet/measurement/gcloud/gcs/GcsStorageOptions.kt` line 24 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/398#-OyKDYI_2ddqxD8cxiPT:-OyKDYI_2ddqxD8cxiPU:b-8z1zs7) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/724a646155d6136ee67bf8d722c2e4eb55c265b6/src/main/kotlin/org/wfanet/measurement/gcloud/gcs/GcsStorageOptions.kt#L24)):*
> ```Kotlin
> 
> /**
>  * Resilience configuration -- retry budget and per-attempt timeouts -- applied to every
> ```

nit: insert newlines to separate summary fragment from paragraphs. See https://developer.android.com/kotlin/style-guide#documentation

Marco-Premier (commented):
> @Marco-Premier partially reviewed 1 file and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/398)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on Marco-Premier).

PR #399, opened 2026-07-24
build!: adopt upstream tink-awskms (AWS SDK v2) and deprecate custom AwsKmsClient

stevenwarejones:
Bumps Tink to 1.23.0 and adopts the upstream `tink-awskms` client (AWS SDK v2). The three AWS KMS factories (`AwsKmsClientFactory`, `GCloudToAwsKmsClientFactory`, `ConfidentialSpaceToAwsKmsClientFactory`) now always return it, wrapped in a new `CompletionExceptionTranslatingKmsClient` (see below). The custom `AwsKmsClient` is not deleted — it stays `@Deprecated` and otherwise unchanged (still Base64-only associated-data encoding, as it always was), but is no longer referenced by any factory. It's kept as a ready-to-wire-back-in fallback in case a currently-undiscovered consumer turns out to need it; if that never materializes it can be deleted outright in a follow-up.

Adds `CompletionExceptionTranslatingKmsClient`, which wraps the upstream client so `encrypt`/`decrypt` translate only a `GeneralSecurityException` cause wrapped in the tracked `CompletionException` leak from a failed credential refresh (tink-crypto/tink-java-awskms#5) into a new `GeneralSecurityException`. A `null`, `RuntimeException`, or `Error` cause means this isn't that specific shape, so the `CompletionException` itself is rethrown unchanged rather than being unwrapped or otherwise guessed at. It is generic over any `KmsClient` and does not attempt any other exception translation: `getAead` calls `delegate.getAead` directly and propagates whatever it throws unchanged, including the known case where upstream throws `IllegalArgumentException` for a malformed key URI instead of `GeneralSecurityException` -- that issue (tink-crypto/tink-java-awskms#9) is intentionally not handled by this wrapper.

[...]

PR #401, opened 2026-07-25
fix!: encode AWS KMS associated data as hex to match Tink

stevenwarejones:
The custom AWS KMS `AwsKmsClient` encoded associated data into the KMS encryption context as Base64, which is NOT interoperable with upstream Tink's `AwsKmsAead` (which uses hex via `BinaryUtils.toHex`). It now encodes as hex unconditionally — byte-for-byte matching Tink — so ciphertexts are interoperable and forward-compatible with a future migration to `tink-awskms`. It supports only what the official Tink client does; the previously proposed configurable Base64/hex mode has been removed.

Standalone and unstacked: no Tink version change, no new runtime dependencies. The longer-term change that adopts upstream `tink-awskms` and bumps Tink to 1.23 is #399.

BREAKING-CHANGE: `AwsKmsClient` now encodes the KMS encryption-context associated data as hex instead of Base64. Data encrypted by a previous version with non-empty associated data cannot be decrypted by this version and must be re-generated.

Issue: world-federation-of-advertisers/cross-media-measurement#4298

### DIFF
--- a/src/main/kotlin/org/wfanet/measurement/aws/kms/AwsKmsClient.kt
+++ b/src/main/kotlin/org/wfanet/measurement/aws/kms/AwsKmsClient.kt
@@ -17,7 +17,6 @@ package org.wfanet.measurement.aws.kms
 import com.google.crypto.tink.Aead
 import com.google.crypto.tink.KmsClient
 import java.security.GeneralSecurityException
-import java.util.Base64
 import java.util.Locale
 import software.amazon.awssdk.auth.credentials.AwsCredentialsProvider
 import software.amazon.awssdk.core.SdkBytes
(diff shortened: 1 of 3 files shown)


SanjayVas (approved):
> @SanjayVas reviewed 3 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/401)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on Marco-Premier).

SanjayVas (approved):
Is there more specific guidance that can be provided to operators to determine what specific actions they may need to take? e.g. how this affects PX or EDPA? Those of course would be part of the BREAKING-CHANGE for the corresponding cmms PR.

> @SanjayVas made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/401#-:-OyZxG3zD6DJNcWnF-Px:b-186qi3)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on Marco-Premier).

PR #402, opened 2026-07-30
feat: add direct Confidential Space to AWS KMS access

Marco-Premier:
Add ConfidentialSpaceToAwsKmsClientFactory + ConfidentialSpaceToAwsWifCredentials and a junixsocket-based ConfidentialSpaceTokenClient that fetches an AWS_PRINCIPALTAGS attestation token from the launcher socket and exchanges it with AWS STS AssumeRoleWithWebIdentity, with no intermediary GCP Workload Identity pool or service account. Enables cross-media-measurement to add a third KMS path for data providers that cannot host a GCP project.

Issue: https://github.com/world-federation-of-advertisers/cross-media-measurement/issues/4309


src/main/kotlin/org/wfanet/measurement/gcloud/confidentialspace/ConfidentialSpaceTokenClient.kt
```diff
@@ -0,0 +1,136 @@
...
+    private fun connectUnixDomainSocket(path: Path): Socket =
+      AFUNIXSocket.connectTo(AFUNIXSocketAddress.of(path.toFile()))
+
+    private fun parseTokenResponse(responseBytes: ByteArray): String {
+      val response = String(responseBytes, StandardCharsets.UTF_8)
+      val separatorIndex = response.indexOf(HEADER_BODY_SEPARATOR)
+      require(separatorIndex >= 0) { "Malformed HTTP response from launcher token socket" }
+      val statusLine = response.substringBefore("\r\n")
+      val statusCode = statusLine.split(" ").getOrNull(1)?.toIntOrNull()
+      val tokenBody = response.substring(separatorIndex + HEADER_BODY_SEPARATOR.length).trim()
```
dawn-wang22: This response parser does not handle `Transfer-Encoding: chunked`. It returns everything after the first `\r\n\r\n` verbatim (only `.trim()`-ed), ignoring `Content-Length`/`Transfer-Encoding` and relying on read-to-EOF + `Connection: close`.

I checked the launcher source (`google/go-tpm-tools`, `launcher/teeserver/tee_server.go` — the `attest` handler this `POST /v1/token` hits): it writes the token with `w.WriteHeader(200); w.Write(tok)` and sets **no `Content-Length`**. Go's `net/http` only auto-sets `Content-Length` when the body fits its internal ~2 KB buffer before the handler returns; larger bodies are sent as `Transfer-Encoding: chunked`. Confidential Space attestation JWTs exceed 2 KB (image-digest/signature claims, GPU device reports, and here the `AWS_PRINCIPALTAGS` claims), so the real response will be chunked.

When that happens, the returned "token" is wrapped in chunk framing (e.g. `1a2f\r\n<jwt>\r\n0`), and STS `AssumeRoleWithWebIdentity` rejects it with `InvalidIdentityToken` — i.e. the path fails for every real token, even though the current unit tests (small, unframed responses) pass.

Suggest either making the request via an HTTP client over the junixsocket connection (junixsocket integrates with `HttpURLConnection`/OkHttp) so framing is handled for you — closest to Google's Go sample, which uses `http.Client` — or explicitly de-chunking / honoring `Content-Length` here.

src/test/kotlin/org/wfanet/measurement/gcloud/confidentialspace/ConfidentialSpaceTokenClientTest.kt
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  private fun clientForServer() =
+    ConfidentialSpaceTokenClient(
+      socketPath = Paths.get("unused-in-test"),
+      socketFactory = { AFUNIXSocket.connectTo(serverAddress) },
+    )
+
+  @Test
+  fun `getToken returns the token body from a 200 response`() {
```
dawn-wang22: Test gap that hides the parsing issue on the token client: both cases here write small, unframed responses, so the read-to-EOF path happens to work. Please add a case where the server responds with `Transfer-Encoding: chunked` (and ideally one with `Content-Length`). A chunked-response test against the current code will fail — which is exactly the signal that is missing, since production tokens are chunked.

dawn-wang22 (commented):
Thanks for this — it mirrors the two-hop `GCloudToAwsKmsClientFactory` cleanly and matches Google's documented token flow. Two comments inline: one on the token client (worth resolving before merge) and one on test coverage. The rest are nits below.

**Nits:**
- `parseTokenResponse` treats only exactly `200` as success; `statusCode !in 200..299` would be marginally more robust. The launcher returns 200 today, so not a real issue.
- `ConfidentialSpaceToAwsKmsClientFactoryTest` only asserts "fails with a bogus config", so real coverage rests almost entirely on the token-client test — noting given the inline test-gap comment.
- `junixsocket-core` showing `"jar": null` / listed under `skipped` in `maven_install.json` is expected (it is a pom aggregator pulling `-common` + `-native-common`) — flagging only so it is not mistaken for a resolution error.

dawn-wang22 (changes requested):

src/main/kotlin/org/wfanet/measurement/gcloud/confidentialspace/ConfidentialSpaceTokenClient.kt
Marco-Premier: Done.

src/test/kotlin/org/wfanet/measurement/gcloud/confidentialspace/ConfidentialSpaceTokenClientTest.kt
Marco-Premier: Done.

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/402)* status: 0 of 12 files reviewed, 2 unresolved discussions (waiting on dawn-wang22).

dawn-wang22 (approved):

dawn-wang22 (approved):
> @dawn-wang22 reviewed 12 files and all commit messages, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/402)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on Marco-Premier).

laureanobrs (approved):
> @laureanobrs reviewed 12 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/402)* status: all files reviewed, 5 unresolved discussions (waiting on Marco-Premier).

MODULE.bazel
```diff
@@ -267,6 +267,10 @@ maven.install(
         "software.amazon.awssdk:kms",
         "software.amazon.awssdk:sts",
 
+        # Confidential Space launcher token socket. Java 11 has no built-in AF_UNIX
+        # support, so junixsocket is used to reach /run/container_launcher/teeserver.sock.
+        "com.kohlschutter.junixsocket:junixsocket-core:2.10.1",
+
```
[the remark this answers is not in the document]
Marco-Premier: Verified, it works

imports/java/com/kohlschutter/junixsocket/BUILD.bazel
```diff
@@ -0,0 +1,6 @@
+package(default_visibility = ["//visibility:public"])
+
+alias(
+    name = "junixsocket",
+    actual = "@maven//:com_kohlschutter_junixsocket_junixsocket_core",
+)
```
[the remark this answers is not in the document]
Marco-Premier: verified, works fine

Marco-Premier (commented):
> @Marco-Premier reviewed 2 files and all commit messages, made 2 comments, and resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/402)* status: all files reviewed, 2 unresolved discussions (waiting on Marco-Premier).

Marco-Premier (commented):
> @Marco-Premier resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/402)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on Marco-Premier).

SanjayVas (commented):
> @SanjayVas made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/402)* status: all files reviewed, 1 unresolved discussion.

___
*[`MODULE.bazel` line 270 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/402#-OzD4XyfBXpAkoh9IJ9F:-OzD4XyfBXpAkoh9IJ9G:bduq8in) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/68c0d48ba6a7abb299ebc866404ca8eccaa8ddfd/MODULE.bazel#L270)):*
> ```Starlark
>         "software.amazon.awssdk:sts",
> 
>         # Confidential Space launcher token socket. Java 11 has no built-in AF_UNIX
> ```

We're already on JDK 17. There's an argument for bumping our Java language version to 17 as well. Barring that, we already have some dependencies on Netty so it would make a more logical choice as the third-party library to use.

PR #403, opened 2026-08-05
feat: use Reactor Netty for the Confidential Space token client

Marco-Premier:
Replaces junixsocket and the hand-rolled HTTP parsing with Reactor Netty over a Netty domain socket, so response framing is handled by the HTTP client rather than by this code.

Credentials are resolved through an `IdentityProvider` of `AwsCredentialsIdentity` rather than an `AwsCredentialsProvider`, so the token fetch and the STS exchange are chained without a `runBlocking` bridge. `AwsKmsClient` takes the wider type; `AwsCredentialsProvider` is a subtype, so its other callers are unaffected. Synchronous clients still join the future inside the SDK, which moves the blocking boundary into `AwsCredentialsAuthorizationStrategy` rather than removing it.

Adds one artifact, `reactor-netty-http`, versioned by the existing `reactor-bom`. `netty-codec-http`, `netty-transport-native-unix-common` and `identity-spi` are declared because they are now used directly, pinned to the versions already resolved so no new jars enter the lockfile. Aligning the Netty artifacts on a single version is tracked separately in #404.

Issue: https://github.com/world-federation-of-advertisers/cross-media-measurement/issues/4309


SanjayVas (changes requested):
> @SanjayVas reviewed 4 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/403)* status: 4 of 8 files reviewed, 2 unresolved discussions (waiting on Marco-Premier).

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/403#-OzIMZRm-LmZzuCHTTUa:-OzIMZRm-LmZzuCHTTUb:bg50mit) (no related file):*
There's a tradeoff that I don't see discussed: The choice of the third-party Netty library vs. `UnixDomainSocketAddress` from the Java standard library.

We generally pick the standard library when possible, so use of another library needs to be justified.
* Use of this class requires updating to Java 17.
   * We already build against JDK 17 and use it in our containers, but currently target Java language level 11.
   * Are there concerns with upgrading the Java language level, e.g. some EDP or MP that cannot use Java 17?
* Is there some other advantage that Netty offers?
   * Commonly the issue is that we want true non-blocking behavior whenever possible (as opposed to just running blocking code in another thread to get it to be "async"). Is that the case here, meaning we cannot get this behavior easily via the Java standard library? `AsynchronousServerSocketChannel` looks like it would work for this.
___
*[`imports/java/io/netty/BUILD.bazel` line 8 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/403#-OzILHe-0to3efkypSLX:-OzILHe-0to3efkypSLY:bbfjv95) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/cba1de18745899e23d11ac803ba6e2eda7cbde4c/imports/java/io/netty/BUILD.bazel#L8)):*
> ```Starlark
> )
> 
> alias(
> ```

Make sure each of the Bazel package paths under //imports/java matches the corresponding Java package path of the library. For example, this netty-codec-http appears to add symbols in multiple Java packages under io.netty.handler.codec. This means it should be defined in `//imports/java/io/netty/handler/codec`. It is not the only artifact which adds symbols under that package, so it should have a name other than the default name for the package (e.g. `http` instead of `codec`).

SanjayVas (changes requested):
> @SanjayVas made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/403)* status: 4 of 8 files reviewed, 2 unresolved discussions (waiting on Marco-Premier).

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/403#-OzIMZRm-LmZzuCHTTUa:-OzIQr1REs9kmlzRQY7Y:b-9xzq08) (no related file):*
> `AsynchronousServerSocketChannel` looks like it would work for this.

Sorry, this is a client so I believe it would just be `java.net.http.HttpClient` with `UnixDomainSocketAddress` as the address.

SanjayVas (changes requested):
> @SanjayVas made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/403)* status: 4 of 8 files reviewed, 2 unresolved discussions (waiting on Marco-Premier).

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/403#-OzIMZRm-LmZzuCHTTUa:-OzITKibB4qx4XNiqsbX:b4mzpi) (no related file):*
> Sorry, this is a client so I believe it would just be `java.net.http.HttpClient` with `UnixDomainSocketAddress` as the address.

Nevermind, `HttpClient.Builder` doesn't have a way to use a `UnixDomainSocketAddress`. It looks like the right choice is likely not using Netty directly, but rather OkHttp (or Ktor with the OkHttp engine). We have dependencies for both OkHttp and Ktor already in the main repo, though I think Ktor is being incorrectly used with the CIO engine in the MCP server despite us already having an OkHttp dependency.

SanjayVas (changes requested):
> @SanjayVas made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/403)* status: 4 of 8 files reviewed, 2 unresolved discussions (waiting on Marco-Premier).

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/403#-OzIMZRm-LmZzuCHTTUa:-OzIWIqw0MRN9WFFR2n4:b-natp7h) (no related file):*

> Sorry, this is a client so I believe it would just be `java.net.http.HttpClient` with `UnixDomainSocketAddress` as the address.

Nevermind, `HttpClient.Builder` doesn't have a way to use a `UnixDomainSocketAddress`. It looks like the right choice is likely not using Netty directly, but rather OkHttp (or Ktor with the OkHttp engine). We have dependencies for both OkHttp and Ktor already in the main repo, though I think Ktor is being incorrectly used with the CIO engine in the MCP server despite us already having an OkHttp dependency.
</blockquote></details>

Sorry for the message spam - I realized I may not have been entirely clear on what I think the options are. Feel free to suggest another if you have a better one.

Basically there are two issues:
1. Unix domain sockets
    1. `UnixDomainSocketAddress` requiring Java 17
    2. Netty's `DomainSocketAddress`
2. Non-blocking HTTP clients so we don't need to manually implement HTTP
    1. OkHttp. Can be used with `UnixDomainSocketAddress` with some adapting.
    2. Ktor
        1. OkHttp engine. Same as (i) but with a more Kotlin-idiomatic interface
        2. Netty engine. Allows use of Netty `DomainSocketAddress`.

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/403)* status: 0 of 17 files reviewed, 2 unresolved discussions (waiting on SanjayVas).

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/403#-OzIMZRm-LmZzuCHTTUa:-OzLnoqh-josdwmMR8i2:b3lgmnp) (no related file):*

Sorry for the message spam - I realized I may not have been entirely clear on what I think the options are. Feel free to suggest another if you have a better one.

Basically there are two issues:
1. Unix domain sockets
    1. `UnixDomainSocketAddress` requiring Java 17
    2. Netty's `DomainSocketAddress`
2. Non-blocking HTTP clients so we don't need to manually implement HTTP
    1. OkHttp. Can be used with `UnixDomainSocketAddress` with some adapting.
    2. Ktor
        1. OkHttp engine. Same as (i) but with a more Kotlin-idiomatic interface
        2. Netty engine. Allows use of Netty `DomainSocketAddress`.
</blockquote></details>

I went with 1.i + 2.i: `UnixDomainSocketAddress` from the standard library, with OkHttp on top.

On the options:

- Ktor + Netty engine isn't available: Netty is a server-only engine in Ktor. There's no `ktor-client-netty` artifact on Maven Central, only `ktor-server-netty`.
- Ktor + OkHttp engine would work, but it's a wrapper over the same OkHttp client, and we'd be adding a Ktor client dependency we don't have today (the main repo only pulls in the Ktor *server* artifacts for the MCP server).
- Using `OkHttp` as a wrapper not to hand-rolling HTTP

As you predicted, OkHttp needs some adapting: it wants a `javax.net.SocketFactory` and the JDK only exposes Unix domain sockets as a `SocketChannel`, so I added `UnixDomainSocket` and `UnixDomainSocketFactory` to bridge the two.

On the Java level I had no real concert, just thought that using `junixsocket-core` for this simple request would have been simpler, that's all. 

I have also put up a PR here if you can take a look: https://github.com/world-federation-of-advertisers/rules_kotlin_jvm/pull/29
___
*[`imports/java/io/netty/BUILD.bazel` line 8 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/403#-OzILHe-0to3efkypSLX:-OzLr8vB-U-CwILqEpBf:bco1lak) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/cba1de18745899e23d11ac803ba6e2eda7cbde4c/imports/java/io/netty/BUILD.bazel#L8)):*

Make sure each of the Bazel package paths under //imports/java matches the corresponding Java package path of the library. For example, this netty-codec-http appears to add symbols in multiple Java packages under io.netty.handler.codec. This means it should be defined in `//imports/java/io/netty/handler/codec`. It is not the only artifact which adds symbols under that package, so it should have a name other than the default name for the package (e.g. `http` instead of `codec`).
</blockquote></details>

doesn't apply anymore

SanjayVas (changes requested):
> @SanjayVas reviewed 9 files and all commit messages, made 3 comments, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/403)* status: 9 of 17 files reviewed, 2 unresolved discussions (waiting on Marco-Premier).

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/403#-OzIMZRm-LmZzuCHTTUa:-OzNGL0rBH-aHmJ6Iul2:byhxusd) (no related file):*
> Ktor + Netty engine isn't available: Netty is a server-only engine in Ktor.

Ah :(

> On the Java level I had no real concert, just thought that using `junixsocket-core` for this simple request would have been simpler, that's all.

What I was looking for was a *discussion* of the different possible options and their tradeoffs. When this kind of thing comes up, it should be covered in a design doc and reviewed *before* any code is written. All code changes should have been paused until that could be resolved. The doc linked in https://github.com/world-federation-of-advertisers/cross-media-measurement/issues/4309 unfortunately only covers the choice of Confidential Space attestation tokens, but does not cover the detailed design (see the Detailed Design section in the [Halo Design Doc template](https://docs.google.com/document/d/1fnKP4qmHXNUmAGezaLWI4pqDISt002V4M3LqK2Y18Uk/template/preview)).

> As you predicted, OkHttp needs some adapting: it wants a `javax.net.SocketFactory`

Ah drat. Looks like OkHttp is actually still doing blocking IO underneath, so it doesn't actually satisfy (2) for being a non-blocking HTTP client. I suppose that makes sense, as the non-blocking IO primitives like `java.nio.SocketChannel` would require a Java 16+ version of OkHttp. Upon closer look I see that we only use OkHttp in tests in the main repo where the use of blocking IO isn't as much of a concern.

My instinct here would be to raise another possible option to use Netty DomainSocketAddress via Reactor. We already depend on `io.projectreactor.netty:reactor-netty-core` for R2DBC since that's how its non-blocking IO is implemented. kotlinx-coroutines has a library with coroutine bindings for Reactor. This eliminates the need to have our own UnixDomainSocket implementation (which we don't want anyway as that's blocking IO). This in turn would mean we don't need to target Java 17 yet.
___
*[`src/main/kotlin/org/wfanet/measurement/common/net/UnixDomainSocket.kt` line 134 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/403#-OzNFLnR8HVjtDgCQBeT:-OzNFLnR8HVjtDgCQBeU:b-55jxrs) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/f252e5ba69efc7c0245fd391e6de4a723be7e691/src/main/kotlin/org/wfanet/measurement/common/net/UnixDomainSocket.kt#L134)):*
> ```Kotlin
>    * Blocks until [selector]'s channel is ready, honoring [soTimeoutMillis].
>    *
>    * Callers depend on a timeout actually firing: OkHttp probes a pooled connection by setting a 1ms
> ```

nit: don't include OkHttp details here
___
*[`src/main/kotlin/org/wfanet/measurement/common/net/UnixDomainSocket.kt` line 186 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/403#-OzNFW8c1O3yeVtnZW2P:-OzNFW8c1O3yeVtnZW2Q:b-jimjls) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/f252e5ba69efc7c0245fd391e6de4a723be7e691/src/main/kotlin/org/wfanet/measurement/common/net/UnixDomainSocket.kt#L186)):*
> ```Kotlin
>   }
> 
>   // A Unix domain socket is addressed by path, so it has no internet address or port. These return
> ```

Document

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/403)* status: 9 of 17 files reviewed, 3 unresolved discussions (waiting on Marco-Premier).

SanjayVas (changes requested):
> @SanjayVas partially reviewed 20 files and all commit messages, made 2 comments, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/403)* status: all files reviewed, 3 unresolved discussions (waiting on Marco-Premier).

___
*[`src/main/kotlin/org/wfanet/measurement/gcloud/confidentialspace/ConfidentialSpaceTokenClient.kt` line 141 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/403#-Ozgfkj6ELitlkqKDi6T:-Ozgfkj7BycA4fBRYcxj:b-yowu9u) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/b157e35eea442ed5a1aa437cc180e8300f3b3ea7/src/main/kotlin/org/wfanet/measurement/gcloud/confidentialspace/ConfidentialSpaceTokenClient.kt#L141)):*
> ```Kotlin
>         }
>         if (request.tokenType == ConfidentialSpaceTokenType.AWS_PRINCIPAL_TAGS) {
>           // For AWS_PRINCIPALTAGS the launcher reads aws_principal_tag_options unconditionally.
> ```

I have no idea what this is talking about. Is there some known issue filed against the corresponding component that you can reference here?

Note that this is assuming it's about something in the component being called by this code. Recall that comments should have the same separation of concerns as code, so this shouldn't reference anything that is about a specific *caller* of this code.
___
*[`src/main/kotlin/org/wfanet/measurement/gcloud/kms/ConfidentialSpaceToAwsKmsClientFactory.kt` line 77 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/403#-OzggPNy2uITkvw6CIU-:-OzggPNy2uITkvw6CIU0:b-4gotkm) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/b157e35eea442ed5a1aa437cc180e8300f3b3ea7/src/main/kotlin/org/wfanet/measurement/gcloud/kms/ConfidentialSpaceToAwsKmsClientFactory.kt#L77)):*
> ```Kotlin
>           "Credentials must not be resolved on a Reactor event-loop thread"
>         }
>         runBlocking { obtainAwsCredentials(config) }
> ```

runBlocking should only be called from the entry point (e.g. `main` or a test method).

You may need to update the interface to make this function `suspend`. This requires updating all the callers as well as updating the other implementations to make sure that any blocking calls happen inside an executor/dispatcher suitable for blocking code.

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/403)* status: 21 of 22 files reviewed, 3 unresolved discussions (waiting on SanjayVas).

___
*[`src/main/kotlin/org/wfanet/measurement/gcloud/confidentialspace/ConfidentialSpaceTokenClient.kt` line 141 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/403#-Ozgfkj6ELitlkqKDi6T:-OzhK2IJ8p6ugFKfPq_6:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/b157e35eea442ed5a1aa437cc180e8300f3b3ea7/src/main/kotlin/org/wfanet/measurement/gcloud/confidentialspace/ConfidentialSpaceTokenClient.kt#L141)):*

I have no idea what this is talking about. Is there some known issue filed against the corresponding component that you can reference here?

Note that this is assuming it's about something in the component being called by this code. Recall that comments should have the same separation of concerns as code, so this shouldn't reference anything that is about a specific *caller* of this code.
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/measurement/gcloud/kms/ConfidentialSpaceToAwsKmsClientFactory.kt` line 77 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/403#-OzggPNy2uITkvw6CIU-:-OzhGcqUC4jglH2re_bY:b56jj54) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/b157e35eea442ed5a1aa437cc180e8300f3b3ea7/src/main/kotlin/org/wfanet/measurement/gcloud/kms/ConfidentialSpaceToAwsKmsClientFactory.kt#L77)):*

runBlocking should only be called from the entry point (e.g. `main` or a test method).

You may need to update the interface to make this function `suspend`. This requires updating all the callers as well as updating the other implementations to make sure that any blocking calls happen inside an executor/dispatcher suitable for blocking code.
</blockquote></details>

On the `runBlocking`: making `getKmsClient` suspend wouldn't remove it, because that method only builds the provider — the AWS SDK invokes the lambda later from the synchronous `AwsCredentialsProvider.resolveCredentials()`, so a bridge is still needed there. The two options that do remove it are (a) make `getToken` non-suspend again, matching what `GCloudToAwsKmsClientFactory` already does with `idTokenCredentials.refresh()`, or (b) refresh credentials proactively in the background so `resolveCredentials()` never re-enters coroutine code. Which would you recommend?  I think to be non-blocking entirely (b) might be the only option?

SanjayVas (changes requested):
> @SanjayVas made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/403)* status: 21 of 22 files reviewed, 3 unresolved discussions (waiting on Marco-Premier).

___
*[`src/main/kotlin/org/wfanet/measurement/gcloud/kms/ConfidentialSpaceToAwsKmsClientFactory.kt` line 77 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/403#-OzggPNy2uITkvw6CIU-:-OzhLF_86QU8cWKLxiEO:b6cc0sh) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/b157e35eea442ed5a1aa437cc180e8300f3b3ea7/src/main/kotlin/org/wfanet/measurement/gcloud/kms/ConfidentialSpaceToAwsKmsClientFactory.kt#L77)):*
<details><summary><i>Previously, Marco-Premier (marcopremier) wrote…</i></summary><blockquote>

On the `runBlocking`: making `getKmsClient` suspend wouldn't remove it, because that method only builds the provider — the AWS SDK invokes the lambda later from the synchronous `AwsCredentialsProvider.resolveCredentials()`, so a bridge is still needed there. The two options that do remove it are (a) make `getToken` non-suspend again, matching what `GCloudToAwsKmsClientFactory` already does with `idTokenCredentials.refresh()`, or (b) refresh credentials proactively in the background so `resolveCredentials()` never re-enters coroutine code. Which would you recommend?  I think to be non-blocking entirely (b) might be the only option?
</blockquote></details>

Ah, I didn't spot that this was the body of the provider.

If I'm reading the code path correctly there's [an overload of `credentialProvider`](https://docs.aws.amazon.com/java/api/latest/software/amazon/awssdk/awscore/client/builder/AwsClientBuilder.html#credentialsProvider(software.amazon.awssdk.identity.spi.IdentityProvider)) in `KmsClientBuilder` which takes in an `IdentityProvider` of `AwsCredentialsIdentity`. This allows you to implement `resolveIdentity` yourself, which returns a CompletableFuture. I think this should allow you to keep everything non-blocking.

Marco-Premier (commented):
> @Marco-Premier made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/403)* status: 15 of 30 files reviewed, 3 unresolved discussions (waiting on SanjayVas).

___
*[`src/main/kotlin/org/wfanet/measurement/gcloud/kms/ConfidentialSpaceToAwsKmsClientFactory.kt` line 77 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/403#-OzggPNy2uITkvw6CIU-:-Ozkbg8_90LeblbxPsxr:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/b157e35eea442ed5a1aa437cc180e8300f3b3ea7/src/main/kotlin/org/wfanet/measurement/gcloud/kms/ConfidentialSpaceToAwsKmsClientFactory.kt#L77)):*

Ah, I didn't spot that this was the body of the provider.

If I'm reading the code path correctly there's [an overload of `credentialProvider`](https://docs.aws.amazon.com/java/api/latest/software/amazon/awssdk/awscore/client/builder/AwsClientBuilder.html#credentialsProvider(software.amazon.awssdk.identity.spi.IdentityProvider)) in `KmsClientBuilder` which takes in an `IdentityProvider` of `AwsCredentialsIdentity`. This allows you to implement `resolveIdentity` yourself, which returns a CompletableFuture. I think this should allow you to keep everything non-blocking.
</blockquote></details>

Done.

SanjayVas (approved):
> @SanjayVas reviewed 15 files and all commit messages, made 1 comment, and resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/403)* status: all files reviewed, 1 unresolved discussion (waiting on Marco-Premier).

___
*[`src/main/kotlin/org/wfanet/measurement/gcloud/confidentialspace/ConfidentialSpaceTokenClient.kt` line 48 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/403#-OzlnwSg-RpBUv_YxrZl:-OzlnwSg-RpBUv_YxrZm:bfd7jj3) ([raw file](https://github.com/world-federation-of-advertisers/common-jvm/blob/9b55bc7efff220a2a849abb019d61e27533fd2fe/src/main/kotlin/org/wfanet/measurement/gcloud/confidentialspace/ConfidentialSpaceTokenClient.kt#L48)):*
> ```Kotlin
>  * Request for a custom Confidential Space attestation token.
>  *
>  * @param audience The audience baked into the token; echoed back in the `aud` claim.
> ```

nit; technically documenting the property is more appropriate than the constructor param for a data class

Marco-Premier (commented):
> @Marco-Premier reviewed 1 file and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/common-jvm/403)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on Marco-Premier).

